PR #5, opened 2022-05-12
Set keycloak login theme

longshuicy:
To test:
1. run the dev script `sh docker-dev.sh up`
2. login to administrator console, find theme
3. select clowder-theme and save
image

go to clowder frontend and try login, you should see an interface like this:
image

I followed tutorial of https://medium.com/keycloak/change-login-theme-in-keycloak-docker-image-55b5fa5ceec4
CSS is adjusted based on this.

### DIFF
--- a/backend/keycloak/clowder-realm-prod.json
+++ b/backend/keycloak/clowder-realm-prod.json
@@ -1258,7 +1258,7 @@
     "auth": "",
     "ssl": ""
   },
-  "loginTheme": "keycloak",
+  "loginTheme": "clowder-theme",
   "eventsEnabled": false,
   "eventsListeners": [
     "jboss-logging"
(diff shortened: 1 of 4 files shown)


backend/docker-dev.sh
```diff
@@ -2,6 +2,9 @@
 if [ $1 == "up" ]
 then
   docker-compose -f docker-compose.dev.yml up -d
+  # need to add volume so we can copy the clowder theme into the right place inside the container
```
lmarini, line 5: why not just add the volume to the docker compose? with this approach if anyone runs the default docker compose to run the full stack it will not include this.

backend/keycloak/realm-export.json
```diff
@@ -33,8 +33,8 @@
   "verifyEmail": false,
   "loginWithEmailAllowed": true,
   "duplicateEmailsAllowed": false,
-  "resetPasswordAllowed": false,
-  "editUsernameAllowed": false,
+  "resetPasswordAllowed": true,
+  "editUsernameAllowed": true,
```
lmarini, line 37: I am not sure we want to allow them to change their username? I think we just want the username to default to their email address?

lmarini (changes requested):

lmarini (approved):

PR #9, opened 2022-05-25
Download dataset

tcnichol:
This pull requests adds download dataset.

Currently this only works on the api route (not front end).

This also just covers the files and one level of folders.

I wanted to get this in and then once its integrated with the front end adding more functionality.


tcnichol:
This one should now work with nested folders. It should be good to review.

max-zilla (approved):

PR #16, opened 2022-06-04
12 create dataset from zip

tcnichol:
The backend endpoint currently works. You can upload a zip and it will create a dataset with folders and subfolders. 

Right now the zip file is unzipped in the working directory. That should probably be changed but not sure what would be best practice. 

Not connected to the front end.


tcnichol:
The back end endpoint works, but there are 2 things that are still needed.

1. This isn't added to the frontend.
2. Right now the zip file is unzipped in the current working directory of the backend, and then the files are uploaded, and once the dataset is created the zip and the unzipped folder are deleted. This does not seem like the best practice, so anyone please comment on better solutions.

max-zilla:
I updated this a little:
- created a shared function to add file to Mongo and Minio, so we don't have duplicated code for uploading file and uploading from dataset as it is somewhat complex
- modified iteration to unzip only one file from archive at at time. i tried to get it fully streamed from request but no luck yet - this is slightly less vulnerable however.

Tested round-robin upload zip -> download zip and it seems to work as intended. Gonna test a couple more edge cases before merge

max-zilla (approved):

PR #15, opened 2022-06-03
delete metadata when file/dataset deleted

max-zilla:
Fix a few bugs related to file/dataset metadata.

PR #17, opened 2022-06-08
replace delete with delete_many

tcnichol:
This fixes the bug with deleting the file. Formerly an error was thrown. this was because .delete was called and it has to be either delete_one or delete_many

I have used delete_many


longshuicy:
This address the comment I had on @max-zilla your PR. Maybe let's get this merged in first? https://github.com/clowder-framework/clowder2/pull/15

PR #27, opened 2022-06-22
24 store file type size in mongodb

tcnichol:
Files uploaded now have the size in bytes stored as well as the content type. Visible on front end.

Not sure if the bytes are readable, many need to change to more readable format.


longshuicy (approved):
looks good to me.

PR #35, opened 2022-06-27
Seeing if node:14-bullseye docker image includes the correct version …

lmarini:
…of glibc for sharp.


lmarini:
Fixed in https://github.com/clowder-framework/clowder2/pull/44

PR #40, opened 2022-06-30
fix delete error

longshuicy:
The backend endpoint returns "deleted":{id}.
On the frontend Im constructing the {"id": {id}} just in case backend return shape will change in the future.


tcnichol (approved):
I checked this one out and it fixes the error. 

I'm marking approved.

PR #48, opened 2022-07-13
47 keycloak needs volume to store user data

tcnichol:
I believe this is now fixed. 

this link on stackoverflow

https://stackoverflow.com/questions/59681041/how-to-take-db-backup-in-keycloak-docker-container/60554189#60554189

had a suggestion which I used and now, when I do docker-dev.sh down and up, the users are still in keycloak. 

If someone else can test to confirm.

Made the change in docker-compose.yml and docker-compose.dev.yml.


longshuicy (approved):
Approved. Tested and user now persists.

PR #56, opened 2022-07-25
ro crate and bagit

tcnichol:
This pull request adds the ro-crate metadata as well as bagit files. 

Currently a metadata file is written for a file only if that file has metadata attached. 

I think that the bagit stuff is implemented right, but could use some feedback on the ro-crate stuff.


max-zilla (approved):

max-zilla:
I fixed one bug during testing with empty datasets - if you dont have any files, it would skip creating one of the manifest files and fail. After that, it all looked good.

PR #58, opened 2022-07-26
39 large files not uploaded correctly

tcnichol:
This should fix it.

Removed settings.MINIO_UPLOAD_CHUNK_SIZE from file.file.read()

The docs seem to indicate that using .read() is okay for large files, before it just read a chunk of bytes and then kept rewriting the file in minio with the same bytes, which is why large files were never uploaded right.

Will test on some larger files (more than a few GB) to be sure.

### DIFF
--- a/backend/app/routers/datasets.py
+++ b/backend/app/routers/datasets.py
@@ -306,7 +306,7 @@ async def save_file(
         # Use unique ID as key for Minio and get initial version ID
         version_id = None
         while content := file.file.read(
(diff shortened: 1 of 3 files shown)


max-zilla:
I moved the different file upload codes to routes/files/add_file_entry() - can you check whether the fix still needs to be applied there? no changes in this current PR

tcnichol:
Max, your fix should do it. I will close.

PR #66, opened 2022-08-07
Register extractor submit file

tcnichol:
I'm going to go ahead and create this as a draft to get feedback.

Extractors are registered using the extractor-info.json. Files and datasets can be submitted to a running extractor. The message sent is as close to what was sent before so that minimal changes are needed for pyclowder2.

Right now extractors don't run since they rely on api endpoints that aren't in clowder2.0. Users also don't have extractor keys. I will create other issues and possibly link them here as well.


backend/app/rabbitmq/heartbeat_listener.py
```diff
@@ -0,0 +1,92 @@
...
+        else:
+            new_extractor = db["extractors"].insert_one(extractor_db.to_mongo())
+            found = db["extractors"].find_one({"_id": new_extractor.inserted_id})
+            extractor_out = ExtractorOut.from_mongo(found)
+            return extractor_out
+
+
+async def main() -> None:
+    # Perform connection
+    connection = await connect("amqp://guest:guest@localhost/")
```
longshuicy, line 66: can we put this `amqp://guest:guest@localhost/` to `config.py`

longshuicy, line 66: is there any need to add credentials? I saw our old code does that
```
credentials = pika.PlainCredentials("guest", "guest")
parameters = pika.ConnectionParameters("localhost", credentials=credentials)
connection = pika.BlockingConnection(parameters)
```

backend/app/rabbitmq/heartbeat_listener_2.py
```diff
@@ -0,0 +1,59 @@
+import pika
```
longshuicy, line 1: sorry I might've missed 2 weeks of discussion. Is this meant for clowder 2? 
Any reason we are using `aio_pika` vs `pika` in different places?

backend/app/routers/datasets.py
```diff
@@ -516,3 +519,44 @@ async def download_dataset(
...
+            body['host'] = 'http://127.0.0.1:8000'
+            body['secretKey'] = 'secretKey'
+            body['retry_count'] = 0
+            body['filename'] = dataset['name']
+            body['id'] = dataset_id
+            body['datasetId'] = dataset_id
+            body['host'] = 'http://127.0.0.1:8000'
+            body['secretKey'] = 'secretKey'
+            body['resource_type'] = 'dataset'
+            body['flags'] = ""
```
longshuicy, line 545: I'm not sure about if we should spell out the secretKey and host here...

backend/app/models/extractors.py
```diff
@@ -1,9 +1,30 @@
...
     version: float = 1.0
+    updated: datetime = Field(default_factory=datetime.utcnow)
+    author: str
+    contributors: List[str] = []
+    contexts: List[dict] = []
+    repository: Union[list[Repository], None] = None
+    external_services: List[str]
+    libraries: List[str] = []
+    bibtex: List[str]
+    maturity: str = "Development"
```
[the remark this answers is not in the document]
tcnichol: This was in the Extractor Model before I started working.

backend/app/rabbitmq/heartbeat_listener.py
```diff
@@ -0,0 +1,92 @@
+import asyncio
+import json
+from aio_pika import ExchangeType, connect
```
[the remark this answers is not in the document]
tcnichol: Added.

backend/app/rabbitmq/heartbeat_listener.py
tcnichol: done

tcnichol: I do not think it is necessary with this library. The URL has the name and password, and this seems to work fine.

backend/app/rabbitmq/heartbeat_listener_2.py
tcnichol: This is something I can change. Part of the issue I had was I was trying to run this as a background process, and I think running in using just pika caused issues.

If it's better, I can try to use pika again, especially since this will be running as a separate script outside of main.

backend/app/routers/datasets.py
tcnichol: I took that out and used token, to match what happens with files.
So far I haven't actually tested dataset extraction, so this may change again later.

backend/app/rabbitmq/heartbeat_listener.py
tcnichol: I use the parameters now in both the (renamed) sync and async heartbeat listeners

tcnichol:
Made a few changes.

Extractor versions are now strings instead of floats, since many extractor versions use 1.0.0. Still compared.

There is also a sync and async listener. Both work, but I would consider the sync one the default and adequate for what we are doing. The async one I wrote because I was trying to get the heartbeat extractor to run as a background thread in main, but if it's run as a separate service it should not matter.

I have not added docker deployment for this. That will be a later issue and pull request.

tcnichol:
Also, for this to work use branch
50-clowder20-submit-file-to-extractor
branch with pyclowder, and test using the wordcount.

tcnichol:
related pull request for pyclowder

branch:

https://github.com/clowder-framework/pyclowder/tree/50-clowder20-submit-file-to-extractor

backend/app/rabbitmq/heartbeat_listener_async.py
```diff
@@ -0,0 +1,85 @@
...
+            existing_version = existing_extractor["version"]
+            new_version = extractor_db.version
+            if version.parse(new_version) > version.parse(existing_version):
+                new_extractor = db["extractors"].insert_one(extractor_db.to_mongo())
+                found = db["extractors"].find_one({"_id": new_extractor.inserted_id})
+                removed = db["extractors"].delete_one(
+                    {"_id": existing_extractor["_id"]}
+                )
+                extractor_out = ExtractorOut.from_mongo(found)
+                print(
```
max-zilla, line 38: let's either remove these print() statements or use logger

backend/app/routers/extractors.py
```diff
@@ -0,0 +1,100 @@
...
+clowder_bucket = os.getenv("MINIO_BUCKET_NAME", "clowder")
+
+
+@router.post("", response_model=ExtractorOut)
+async def save_extractor(
+    extractor_in: ExtractorIn,
+    user=Depends(keycloak_auth.get_current_user),
+    db: MongoClient = Depends(dependencies.get_db),
+):
+    result = extractor_in.dict()
```
max-zilla, line 28: unused

max-zilla (changes requested):
I think this is mostly good for initial implementation, just a few changes.

backend/app/models/extractors.py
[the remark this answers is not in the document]
lmarini: It's to identify extractors that are still under development versus those that are ready for production.

We had a discussion about this and decided to keep fields in and review all of them together.

max-zilla (approved):

PR #80, opened 2022-09-02
Metadata4more resource types

MBcode:
While getting started my 1st user did not have a first/last name, so login was blocked, 
So I changed the access of userinfo to a .get with a default of " " 
Which allowed login, for when you don't have those (I've assumed) not required fields

I also added a changelog for the backend, because I didn't see one yet

ps=I plan to continue to use this branch, to try out at a root resource class, 
   w/the ability to add metadata, so we don't have to replicate those methods


MBcode:
change readme to instructions that allowed seeing altered index
So with 2 changed Readme lines, could add the 1 new index line
 and see the pycharm.md files I had added.
Seemed small but, I should have had in another branch, sorry abou that

backend/CHANGELOG.md
```diff
@@ -0,0 +1,12 @@
...
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/)
+and this project adheres to [Semantic Versioning](http://semver.org/).
+
+
+## [0.0.1] - 2021-11-10
+
+### Added
+- Allow for missing first/last names in userinfo, defaulting to " "
```
longshuicy, line 12: are we doing the changelog now?

PR #81, opened 2022-09-02
57 file does not upload to folder

longshuicy:
Now you can post file inside a folder

For some reason I cannot get folder_id as a part of the form; so instead i just make it a query parameter, hopefully it's ok?


tcnichol (approved):
I tested this on a folder, subfolder, and a subsubfolder. Appears to work fine marking approved.

PR #83, opened 2022-09-02
fix content type

longshuicy:


backend/app/routers/files.py
```diff
@@ -47,7 +47,7 @@ async def add_file_entry(
     new_file_id = new_file.inserted_id
     if content_type is None:
         content_type = mimetypes.guess_type(file_db.name)
-    content_type = content_type[0] if len(content_type) > 1 else content_type
+        content_type = content_type[0] if len(content_type) > 1 else content_type
```
longshuicy, line 50: we should use content type as it is if it's being passed in; only extract when guess from filename

tcnichol (approved):
Marking approved for this. Works.

max-zilla (approved):

PR #92, opened 2022-09-08
84 frontend implement delete folders

longshuicy:
Frontend implementation is finished
However need to take closer look at the backend DELETE folder endpoint:
- two bugfix PRs are needed: https://github.com/clowder-framework/clowder2/pull/85 and https://github.com/clowder-framework/clowder2/pull/81
- investigate if it also deletes any child folders, add that feature


longshuicy:
Ready to be reviewed. Implemented recursively delete folder

lmarini (approved):
Work as intended both from the GUI as well as the API endpoint. 
I didn't not look closely at the code.

PR #98, opened 2022-09-21
Fix/pytest keycloak timeout

lmarini:
I had to change keycloak back to not mounting the local data directory use the postgres services. The env variables changed with the new version of Keycloak and it wasn't connecting to it at all. It was default to H2 database. @tcnichol the change you made fixed it because it was then storing the H2 files in a volume, but it was still ignoring the posgres database. I am not sure why this fixed the pytest build on the server.


lmarini:
@ddey2 the elasticsearch test claims it passed, but in the shell there is an error printed out. Not sure why.

> PASSED                       [ 36%]ERROR:app.elastic_search.connect:BadRequestError(400, 'illegal_argument_exception', 'unknown setting [index.mappings.properties.creator.type] please check that any required plugins are installed, or check the breaking changes documentation for removed settings')

ddey2:
@lmarini I will checkout this branch and try this out. I will keep you posted.

max-zilla (approved):
Verified this fixed the tests.

PR #106, opened 2022-09-26
Directories cleanup

lmarini:
Rearranged directories so that docs, keycloak, scripts are not in backend anymore and
are easier to find. Moved some files around as well. Top level directories are now:
/.github
/backend
/docker
/docs
/frontend
/scripts


longshuicy:
I'm wondering is it possible to change "keycloak" to "clowder_keycloak"; as well as "keycloak.py" to "clowder_keycloak.py" to avoid confusion when important and potential naming collision? We are installing the python keycloak package it's also named "keycloak".

But ^ this doesn't break anything so far so it's not very necessary at least for now.

longshuicy:
PR looks good. Approving.

longshuicy (approved):

lmarini:
> I'm wondering is it possible to change "keycloak" to "clowder_keycloak"; as well as "keycloak.py" to "clowder_keycloak.py" to avoid confusion when important and potential naming collision? We are installing the python keycloak package it's also named "keycloak".
> 
> But ^ this doesn't break anything so far so it's not very necessary at least for now.

The `/scripts/keylcloak` folder? It's not a python folder so it shouldn't be an issue?

Or the `/backend/app/routers/keycloak.py` module? maybe `keycloak_routes.py`?  or `kc.py`? I would rather avoid repeating the word clowder.

longshuicy:
> > I'm wondering is it possible to change "keycloak" to "clowder_keycloak"; as well as "keycloak.py" to "clowder_keycloak.py" to avoid confusion when important and potential naming collision? We are installing the python keycloak package it's also named "keycloak".
> > But ^ this doesn't break anything so far so it's not very necessary at least for now.
> 
> The `/scripts/keylcloak` folder? It's not a python folder so it shouldn't be an issue?
> 
> Or the `/backend/app/routers/keycloak.py` module? maybe `keycloak_routes.py`? or `kc.py`? I would rather avoid repeating the word clowder.

maybe just `keycloak.py` to `keycloak_routes.py`

PR #108, opened 2022-09-26
Explore page cleanup

lmarini:
- Removed placeholder right column on Explore page.
- Switched to clickable card for datasets.
- Added action buttons at bottom.
- Fixed layout so that cards don't fill out the page width.


frontend/src/components/Dashbard.tsx
```diff
@@ -127,14 +130,17 @@ export const Dashboard = (): JSX.Element => {
...
 				<ActionModal actionOpen={errorOpen} actionTitle="Something went wrong..." actionText={reason}
 							 actionBtnName="Report" handleActionBtnClick={handleErrorReport}
 							 handleActionCancel={handleErrorCancel}/>
+				<Box m={1} display="flex" justifyContent="space-between" alignItems="flex-end">
+					<MainBreadcrumbs paths={paths}/>
+					<Button href="/create-dataset" variant="contained" sx={{display: "flex", alignItems: "center"}}>New
+						Dataset</Button></Box>
 				<div className="inner-container">
 					<Grid container spacing={4}>
-						<Grid item xs={12} md={8}>
```
longshuicy, line 137: I think the xs, md, lg, xl might be needed on the grid. otherwise it looks weird with small number of items

lmarini: Thanks for catching this. Should be fixed now

longshuicy (approved):

PR #114, opened 2022-09-30
Extractors -> listeners, trigger listeners on saved feeds automatically

max-zilla:
This renames extractors to listeners (and creates a legacy endpoint for old extractors) as well as includes basic implementation of job feeds (saved searches) and listeners on those feeds. Lots of decisions here we should discuss.

See test_feeds.py for implementation example, still in progress and very rough.


tcnichol (approved):
Ran this one and it still works with extractors. Looks good. Marking approved.

tcnichol:
I think the failing test might be due to ListenerIn needing to be a superclass of ExtractorInfo instead of BaseModel, or perhaps the other way around.

lmarini:
Will you be adding https://github.com/clowder-framework/clowder2/blob/workflow-documentation/backend/docs/source/listeners.md to this PR?

lmarini:
Are you still in favor or renaming `Listeners` to `EventListeners` to be a bit more specific. When looking through the code and documentation I am still wondering if `Listeners` is still too generic, even for a developer.

max-zilla:
> Can you add docstrings to all functions and models? We need to start requiring this on all PRs. Thanks!

Added a bunch of these.

> Will you be adding https://github.com/clowder-framework/clowder2/blob/workflow-documentation/backend/docs/source/listeners.md to this PR?

Added with minor updates as well.

> Are you still in favor or renaming `Listeners` to `EventListeners` to be a bit more specific. When looking through the code and documentation I am still wondering if `Listeners` is still too generic, even for a developer.

I renamed to EventListeners across the app, although "listeners.py" is still the filename for simplicity.

backend/app/models/listeners.py
```diff
@@ -0,0 +1,88 @@
...
+    repository: List[Repository] = []
+    external_services: List[str] = []
+    libraries: List[str] = []
+    bibtex: List[str] = []
+    default_labels: List[str] = []
+    categories: List[str] = []
+    parameters: List[dict] = []
+
+
+class EventListenerBase(BaseModel):
```
lmarini, line 33: Is this supposed to inherit from `ExtractorInfo`?

backend/app/models/listeners.py
```diff
@@ -0,0 +1,88 @@
...
+    """On submission, minimum info for a listener is name, version and description. Clowder will use name and version to locate queue."""
+
+    pass
+
+
+class LegacyEventListenerIn(ExtractorInfo):
+    """v1 Extractors can submit data formatted as a LegacyEventListener (i.e. v1 format) and it will be converted to a v2 EventListener."""
+
+    name: str
+    version: str = "1.0"
```
lmarini, line 51: Why the difference in types between version for `EventListenerBase` and `LegacyEventListener` (str vs int)?

backend/app/tests/test_feeds.py
```diff
@@ -0,0 +1,70 @@
...
+    file_data = {"file": open(dummy_file, "rb")}
+    response = client.post(
+        f"{settings.API_V2_STR}/datasets/{dataset_id}/files",
+        headers=headers,
+        files=file_data,
+    )
+    os.remove(dummy_file)
+    assert response.status_code == 200
+
+    # Verify the message
```
lmarini, line 70: Is there a way to test that the listener was triggered?

backend/app/routers/feeds.py
```diff
@@ -0,0 +1,158 @@
...
+    FeedDB,
+    FeedOut,
+)
+from app.models.search import SearchIndexContents
+from app.search.connect import check_search_result
+from app.rabbitmq.listeners import submit_file_message
+
+router = APIRouter()
+
+clowder_bucket = os.getenv("MINIO_BUCKET_NAME", "clowder")
```
lmarini, line 26: I don't think this is being used.

backend/app/routers/feeds.py
```diff
@@ -0,0 +1,158 @@
...
+        if found_auto:
+            # Verify whether resource_id is found when searching the specified criteria
+            feed_match = check_search_result(es_client, file_out, feed_db.search)
+            if feed_match:
+                for listener in feed_db.listeners:
+                    if listener.automatic:
+                        listeners_found.append(listener.listener_id)
+
+    for targ_listener in listeners_found:
+        queue = ""  # TODO: Each extractor gets a queue - routing key same as name?
```
lmarini, line 75: Can this default to the listener name so that legacy extractors can be triggered?

lmarini (changes requested):
Thanks for updating the PR. Left a few more questions / comments as I continue to review this. There is a lot of digest. Thanks!

backend/app/models/listeners.py
max-zilla: Clowder v1 lets users arbitrarily provide a version with no rules (e.g. "1", "1.0", "v1" "beta" all allowed). I was thinking (before v1 back-compatibility) that version could just be an integer for clarity and simplicity, but after seeing it is just a string field for v1 we should probably use that. 

If we use string, "version" might better be called "tag" or something like Docker does? Don't necessarily see a point in enforcing something that can be cast as a float in that case...

lmarini: > Clowder v1 lets users arbitrarily provide a version with no rules (e.g. "1", "1.0", "v1" "beta" all allowed). I was thinking (before v1 back-compatibility) that version could just be an integer for clarity and simplicity, but after seeing it is just a string field for v1 we should probably use that.
> 
> If we use string, "version" might better be called "tag" or something like Docker does? Don't necessarily see a point in enforcing something that can be cast as a float in that case...

Good point. Tag sounds great. We could also enforce semantic versioning in the string and call it version. But that might be more work. Similar to what npm does.

backend/app/tests/test_feeds.py
max-zilla: not yet. if we had an event log we could check that but it doesn't exist yet. Otherwise we'd have to check RMQ somehow.

backend/app/models/listeners.py
lmarini: Sounds good. We can try to keep both. Maybe add a few more comments explaining the difference? As long as we can submit/trigger with current extractors, we are good short term. I don't think this gets in the way of that?

backend/app/tests/test_feeds.py
lmarini: @tcnichol how easily could you add v1 events back into v2? I am looking at the heartbeat PR and will leave comments there soon. It would be a similar approach?

lmarini (approved):

PR #116, opened 2022-09-30
Kubernetes

lmarini:
Initial files created with kompose. minio-nginx, frontend, backend, keycloak containers failing.


robkooper:
I think we should use a helm chart, I will create a draft version of it.

robkooper:
see #119

lmarini:
Replaced by #119

PR #123, opened 2022-10-03
[WIP] 122 minify keycloak realm json so kubernetes clusters have easier way to load it

longshuicy:
I tested the `clowder-realm-dev.json` on my local machine by just doing `sh docker-dev.sh up`; it's using the docker-compose.dev.yaml where mount this json volume... seems ok to me? But not sure is there other ways I can test this


scripts/keycloak/clowder-realm-dev.json
```diff
@@ -1,282 +1,11 @@
...
-      ],
-      "optionalClientScopes": [
-        "address",
-        "phone",
-        "offline_access",
-        "microprofile-jwt"
-      ]
-    },
-    {
-      "id": "9d3fa7e0-6aa2-45c1-865d-4eee673eb6bf",
```
robkooper, line 279: do we need all the default values here? or can we remove more?

robkooper (approved):
I think this cleans it up a bunch, just wondering if we can remove more in the client and identityProviders sections. I think most of those are default values as well

lmarini (approved):

PR #128, opened 2022-10-06
127 add parameters to extractor submit add submit dataset

tcnichol:
This pull request does the following:

It is now possible to submit datasets to extractors.
Parameters can be added for either file or dataset extractors. 

This relies on a pull request from pyclowder :

That branch has changes to make it work with the new API. If extractors are run using a different pyclowder, they will not work.


lmarini (changes requested):
Can you add comments, doc strings, and pytest? Thank you.

tcnichol:
Will do.

tcnichol:
Testing is implemented and build errors there fixed.

lmarini:
Parent branch merged into main. Please update to resolve conflicts. Thank you.

tcnichol:
I'm getting errors with the pytest because the extractor_info has an author field (a string) but Listeners have an Author that is UserOut.

Could we perhaps use a different term (maybe 'Creator') for the Listener, so as not to conflict with the field in the existing extractor_info.json files?

tcnichol:
New change - 

to make Listeners work when being registered through the heartbeat listener I changed 'author' to 'creator' but also made creator optional. This seemed like the best solution. It also seems to work if Listeners are built in top of Extractors, like someone wants to fine tune when an existing extractor runs. This one should be ready for review again. Let me know if anything else needs to be changed.

max-zilla (approved):

PR #129, opened 2022-10-10
Drawer menu

lmarini:
**Depends on #108 . Please merge that first.**

New Layout component to wrap pages using the MUI Drawer. All new pages should include this Layout.

Fixed bug in link to registration.

*TODO: store open/closed variable for drawer in redux store. Right now the drawer always shows closed after switching pages.*
 
Screen Shot 2022-10-10 at 11 58 20 AM


frontend/src/routes.tsx
```diff
@@ -15,6 +15,9 @@ import {isAuthorized} from "./utils/common";
 import {useDispatch, useSelector} from "react-redux";
 import {RootState} from "./types/data";
 import {resetLogout} from "./actions/common";
+import Layout from "./components/Layout";
+import {Paragraph} from "./components/Paragraph";
```
longshuicy, line 19: I don't think Paragraph is being used

longshuicy: Same as Layout

frontend/src/components/Dashbard.tsx
```diff
@@ -140,7 +140,7 @@ export const Dashboard = (): JSX.Element => {
 						Dataset</Button></Box>
 				<div className="inner-container">
 					<Grid container spacing={4}>
-						<Grid item xs>
+						<Grid item>
```
longshuicy, line 143: I still saw the same issue
image

I think last time you tried xs to fix it; but this time it won't work for me

longshuicy (commented):
At any appropriate time, maybe we can run reformat code to get rid of the unused import? 

image

frontend/src/routes.tsx
lmarini: Done. Thanks for the reminder. I try to do it when I remember, but we don't have an automatic way to do this. Also many of my files show in red because of the `import React, {useEffect, useState} from "react";` import at the top. React is never used and Typescript doesn't like that.

frontend/src/components/Dashbard.tsx
lmarini: I made a new file called Explore.tsx to make it clear that that is the explore page. I fixed it there and remove Dasboard.tsx file to not create confusion.

longshuicy (approved):

PR #136, opened 2022-10-12
63 google dataset search support

MBcode:
early functions and route for getting /summary.jsonld 
that can be inserted in the dataset pages for crawling via reading 
a /sitemap.xml route, but will need to be able to make those urls public 1st

### DIFF
--- a/backend/app/routers/datasets.py
+++ b/backend/app/routers/datasets.py
@@ -191,6 +191,40 @@ async def save_dataset(
     return dataset_out
 
 
+def is_str(v):
+    return type(v) is str
+
+schemaOrg_mapping = {
+    "id": "identifier",
+    "first_name": "givenName",
+    "last_name": "familyName",
+    "created": "dateCreated",
+    "modified": "dateModified",
+    "views": "interactionStatistic",
+    "downloads": "DataDownload"
+}
+def datasetout_str2jsonld(jstr):
+    if not is_str(jstr):
+        jt=type(jstr)
+        print(f'str2jsonld:{jstr},wrong type:{jt}')
+        return None
+    global schemaOrg_mapping
+    for k,v in schemaOrg_mapping.items():
+        if k in jstr:
+            ks=f'"{k}":'
+            vs=f'"{v}":'
+            print(f'replace:{ks} with:{vs}')
+            jstr = jstr.replace(ks,vs)
+    #print(f'==jstr:{jstr}')
+    jstr = jstr.replace("{",'{"@context": {"@vocab": "https://schema.org/"},',1)
+    #print(f'==jstr:{jstr}')
+    return jstr
+
+def datasetout2jsonld(dso):
+    jstr=dso.json()
+    return datasetout_str2jsonld(jstr)
+
+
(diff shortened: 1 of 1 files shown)


backend/app/main.py
```diff
@@ -116,6 +116,16 @@
     tags=["extractors"],
     dependencies=[Depends(get_current_username)],
 )
+api_router.include_router(
+    extractors.router,
```
longshuicy, line 120: You newly added routes are in routers/datasets.py; but this new prefix is importing routes from extractors.
So i can actually access your routes under the `/dataset/{id}/summary.jsonld` but not here

I'd recommend you separate them out in a new router file (e.g. summary.py) and include them here. `summary.router`

backend/app/main.py
```diff
@@ -116,6 +116,16 @@
...
+    prefix="/summary.jsonld",
```
longshuicy, line 121: prefix is for the prefix :-p

e.g. prefix = "/dataset" means the routes will all start with /dataset s--> /datasets/{id}, /datasets/{id}/files etc

backend/app/routers/datasets.py
```diff
@@ -226,6 +270,56 @@ async def get_dataset(dataset_id: str, db: MongoClient = Depends(dependencies.ge
     raise HTTPException(status_code=404, detail=f"Dataset {dataset_id} not found")
 
 
+@router.get("/{dataset_id}/summary.jsonld", response_model=DatasetOut)
```
longshuicy, line 273: since this is under dataset, i can actually set break points and hit this route
image
image


But then it errors out and you should investigate
image

backend/app/routers/datasets.py
```diff
@@ -226,6 +270,56 @@ async def get_dataset(dataset_id: str, db: MongoClient = Depends(dependencies.ge
...
+# now the route
+
+
+def get_txtfile(fn):
+    "ret str from file"
+    with open(fn, "r") as f:
+        return f.read()
+
+
+@router.get("/sitemap.xml")
```
longshuicy, line 314: this does not work... when i try `/datasets/sitemap.xml` it try to cast sitemap.xml into an objectId, I'm guessing because it collides with the `/datasets/{id}` endpoint

if you separate these out to separate file and give different prefix than dataset, it might work

backend/app/main.py
MBcode: that was left over, thanks, will skip the tags for now, and take your other advice to put those routes in another file

backend/app/routers/datasets.py
MBcode: separated out into a sitemap.py file, &will make another pass soon, to see if it will open up the route

backend/app/routers/datasets.py
MBcode: during a cleanup I make sure it is a DatasetOut to safe to call .json on it now

backend/app/routers/datasets.py
MBcode: different file now, w/no overlapping routes

PR #139, opened 2022-10-13
Added endpoint to search both file and dataset.

ddey2:


backend/app/routers/elasticsearch.py
```diff
@@ -23,3 +23,10 @@ async def search_dataset(request: Request):
     es = connect_elasticsearch()
     query = await request.body()
     return search_index(es, "dataset", query)
+
+
+@router.post("/_msearch")
+async def search_file_and_dataset(request: Request):
```
longshuicy, line 29: can you rename this route `@router.post("/file,dataset/_msearch")`; that's what reactive search needed

longshuicy (approved):
just left one comment about route; 
because in reactivesearch you have to specify a baseapp name. so it has to be something like /xxxx/_msearch. 

The Documentation suggesting use a comma separate indexes so in this case it's /file,dataset/_msearch

longshuicy:
In addition, could we add more fields to file
I would  definitely need 
```
dataset_id: PyObjectId
folder_id: Optional[PyObjectId]
```

it would be also nice to have  
```
bytes: int = 0
content_type: str = "N/A"
```

backend/app/routers/elasticsearch.py
ddey2: Done

ddey2:
Sure. I can add those fields. Also, if you need anything else, let me know. I will create a separate PR for adding the fields.

longshuicy:
> Sure. I can add those fields. Also, if you need anything else, let me know. I will create a separate PR for adding the fields.

Actually do you mind I create those additional fields in the PR i'm working on? that might be more efficient

ddey2:
@longshuicy Feel free to add those fields.

PR #140, opened 2022-10-13
Updated elasticsearch indices on updating files and datasets

ddey2:


tcnichol:
I ran this with datasets and saw that the indices updated correctly. I also did 'update file' and uploaded a new version of the same file with the same name. Looks like it connects to elasticsearch and all indices get updated.

tcnichol (approved):
I'm going to mark this approved.

It might be nice if clowder tries to reconnect to elasticsearch if it didn't make the initial connection, but that's probably an issue for another pull request so I put this one down approved. Tested for files and datasets.

backend/app/routers/files.py
```diff
@@ -170,6 +179,16 @@ async def update_file(
             creator=user,
         )
         await db["file_versions"].insert_one(new_version.to_mongo())
+        # Update entry to the file index
+        doc = {
+            "doc": {
+                "name": updated_file.name,
+                "creator": updated_file.creator.email,
+                "created": datetime.utcnow(),
+                "download": updated_file.downloads,
```
longshuicy, line 188: I added a few more required fields for the files in another PR. This might be updated...

longshuicy: we can merge this first and then i'll add more fields in my other PR. Looks like this is a smaller PR so it might be easier to be merged in first

lmarini (approved):
We can probably merge this as is, but I just realized that we are not using depends for elasticsearch the same way we do for other service. This would simplify the code, and we might be able to reuse the elasticsearch client instead of creating an instance every time.

For examples of depends see for example `db: MongoClient = Depends(dependencies.get_db)` in the function definitions.

You can learn more about dependency injection in fastapi here https://fastapi.tiangolo.com/tutorial/dependencies/.

@ddey2 can you create an issue for this and start working on it on a different branch? Thank you.

ddey2:
I will create an issue and address this. I am copy pasting your comment there just to keep track.

PR #150, opened 2022-10-19
Implemented Metadata Definition UI

arunapa:
Screen Shot 2022-10-19 at 10 10 33 AM
Screen Shot 2022-10-19 at 10 11 37 AM


longshuicy:
I got a 422 error when posting definition:
image

Comparing your post body with the examples:
1. the context field should have another layer with the name of the definition as the key
2. you are missing the "name" of the definition, e.g. DOI at the root level
```
{
   "metadataName":"doi",
   "metadataDescription":"doi",
   "context":"https://schema.org/doi",
   "fields":[
      {
         "name":"doi",
         "list":false,
         "widgetType":"TextField",
         "config":{
            "type":"str",
            "options":""
         },
         "required":true
      }
   ]
}
```
vs
```
{
    "name" : "DOI",
    "description" : "Digital object identifier",
    "context" : {
        "doi" : "https://schema.org/DigitalDocument"
    },
    "fields" : [
        {
            "name" : "doi",
            "list" : false,
            "widgetType": "TextField",
            "config": {
                "type" : "str"
            },
            "required" : true
        }
    ]
}
```

frontend/src/components/Layout.tsx
```diff
@@ -168,6 +168,17 @@ export default function PersistentDrawerLeft(props) {
 						</ListItemButton>
 					</ListItem>
 				</List>
+				<Divider/>
+				<List>
+					<ListItem key={"newmetadata"} disablePadding>
+						<ListItemButton component={RouterLink} to="/new-metadata-definition">
+							<ListItemIcon>
+								<AddBox/>
```
longshuicy, line 176: I would change to another ICON just to differentiate with the new dataset icon. Maybe this: `import CreateIcon from '@material-ui/icons/Create';`?

frontend/src/components/datasets/Dataset.tsx
```diff
@@ -276,6 +278,9 @@ export const Dataset = (): JSX.Element => {
 											>
 												<CloseIcon />
 											</IconButton>
+											<Button variant="contained" onClick={() => {setOpenPopup(true);}} sx={{ mt: 1, mr: 1, "alignItems": "right" }}>
+												Add new field
```
longshuicy, line 282: Add new metadata definition? Field is confusing to me.
Others, thoughts?

frontend/src/components/datasets/Dataset.tsx
```diff
@@ -286,6 +291,22 @@ export const Dataset = (): JSX.Element => {
 													sx={{ mt: 1, mr: 1 }}>
 												Cancel
 											</Button>
+											{
+												openPopup ?
+												<>
+													<Dialog open={openPopup} onClose={() => {setOpenPopup(false);}}>
+														<DialogTitle>Add a new field</DialogTitle>
```
longshuicy, line 298: Same here about the field

frontend/src/components/metadata/CreateMetadataDefinition.tsx
```diff
@@ -0,0 +1,307 @@
...
+								variant="contained"
+								onClick={postMetadata}
+								sx={{ mt: 1, mr: 1, "alignItems": "right" }}>
+								Submit
+							</Button>
+							<Button
+								onClick={clearForm}
+								sx={{ mt: 1, mr: 1, "alignItems": "right" }}>
+								Clear
+							</Button>
```
longshuicy, line 300: IMO it would be nice to add a "Back" at the very last step too. Sometimes I still want to change things until the very last...

frontend/src/components/metadata/CreateMetadataDefinition.tsx
```diff
@@ -0,0 +1,307 @@
...
+										variant="outlined"
+										margin="normal"
+										required
+										fullWidth
+										id="datatype"
+										label="Field Data Type"
+										placeholder="Please enter field data type (e.g. int, str, enum)"
+										name="Field Data Type"
+										value={input.config.type}
+										onChange={(event) => { handleInputChange(index, "type", event.target.value); }}
```
longshuicy, line 265: Since this right now is also an arbitrary list, could you also make it a dropdown?
```
"int": int,
"float": float,
"str": str,
"TextField": str,
"bool": bool,
"date": datetime.date,
"time": datetime.time,
"dict": dict,  # TODO: does this work?
"enum": str,  # TODO: need a way to validate enum,
"tuple": tuple,
```

longshuicy: For both the widget type and data type drop downs, can we use more layman's term? e.g. TextField --> Text Box; Select --> Dropdown, Int --> Integer, Float --> Number
You are welcome to improvise a little bit :D

frontend/src/components/metadata/CreateMetadataDefinition.tsx
```diff
@@ -0,0 +1,307 @@
...
+									/>
+									{(input.config.type == "enum") ? <>
+										<TextField
```
longshuicy, line 268: If you plan to address the multiple options of Enum in the next PR please create a task and add a TODO here :-)

frontend/src/components/metadata/CreateMetadataDefinition.tsx
```diff
@@ -0,0 +1,307 @@
...
+											<IconButton color="primary" size="small" onClick={() => addNewField(index)}>
+												<AddBoxIcon />
+											</IconButton>
+											<IconButton color="primary" size="small" onClick={() => removeField(index)}>
+												<DeleteOutlineIcon />
+											</IconButton>
+										</>
+										: 
+										<IconButton size="small" onClick={() => removeField(index)}>
+											<DeleteOutlineIcon />
```
longshuicy, line 210: When I click this delete Icon in this state, it behaves very weird
image

longshuicy (changes requested):

frontend/src/components/Layout.tsx
arunapa: Updated
Screen Shot 2022-10-27 at 1 19 01 AM

frontend/src/components/metadata/CreateMetadataDefinition.tsx
arunapa: I've updated the stepper code, now you can click directly on the stepper number to navigate back. I figured that a dding too many buttons would make the UI cluttered, hence adopted this approach instead :)

frontend/src/components/metadata/CreateMetadataDefinition.tsx
arunapa: I've added human readable names in both drop downs

image

image

frontend/src/components/metadata/CreateMetadataDefinition.tsx
arunapa: Added To-dos wherever applicable

frontend/src/components/metadata/CreateMetadataDefinition.tsx
arunapa: Fixed the issue -- the stepper state wasn't being properly updated when an entry was deleted.

arunapa:
Some more improvements I've added in the latest PR:
1. Sanity checks while moving between stepper states (previously sanity was executed prior to invoking the API)
2. Added a new "preview" text box so that the user can know what metadata definition is being sent to the backend 
image

3. Clicking on the stepper can move between states now

longshuicy (approved):

arunapa:
> 

I've addressed the following:
1. Making the capitalization consistent
2. Changing the default context example
3. Making the context example editable for the user to modify quickly while submitting the form
4. Checkboxes in same row

Screen Shot 2022-11-01 at 10 50 38 AM
Screen Shot 2022-11-01 at 10 50 44 AM

lmarini (approved):

PR #152, opened 2022-10-21
Generic error catching in DB

max-zilla:
Adds a function to catch exceptions in the DB with some extra info.


backend/app/search/connect.py
```diff
@@ -15,10 +17,13 @@ def connect_elasticsearch():
     _es = None
     logger.info(settings.elasticsearch_url)
     _es = Elasticsearch(settings.elasticsearch_url)
-    if _es.ping():
-        logger.info("Successfully connected to Elasticsearch")
-    else:
-        logger.info("Can not connect to Elasticsearch")
+    try:
```
lmarini, line 20: Code looks good, but I am not sure how to test it in this case. If I turn off elasticsearch, I get an error before we get to ping.

```elastic_transport.ConnectionError: Connection error caused by: NewConnectionError(<urllib3.connection.HTTPConnection object at 0x10dce7b20>: Failed to establish a new connection: [Errno 61] Connection refused)
WARNING:elastic_transport.node_pool:Node <Urllib3HttpNode(http://localhost:9200)> has failed for 11 times in a row, putting on 30 second timeout
WARNING:elastic_transport.transport:Retrying request after failure (attempt 2 of 3)
Traceback (most recent call last):
  File "/Users/lmarini/.local/share/virtualenvs/clowder2-pkRZ1537/lib/python3.9/site-packages/elastic_transport/_transport.py", line 329, in perform_request
    meta, raw_data = node.perform_request(
  File "/Users/lmarini/.local/share/virtualenvs/clowder2-pkRZ1537/lib/python3.9/site-packages/elastic_transport/_node/_http_urllib3.py", line 199, in perform_request
    raise err from None
elastic_transport.ConnectionError: Connection error caused by: NewConnectionError(<urllib3.connection.HTTPConnection object at 0x10dce7910>: Failed to establish a new connection: [Errno 61] Connection refused)
WARNING:elastic_transport.node_pool:Node <Urllib3HttpNode(http://localhost:9200)> has failed for 12 times in a row, putting on 30 second timeout```

**Review by lmarini — APPROVED** [review]

File: backend/app/search/connect.py (hunk 1 above)
Comment (max-zilla) [reply backend/app/search/connect.py:20]: so this was 1) start up docker ES, 2) start up backend, 3) kill docker ES?

otherwise, we could come up with a contrived test case like  "if es.ping() =.TRUE, pretend it failed" just to check the actual logging of errors even if the error is nonsense.

File: backend/app/search/connect.py [hunk 2]
```diff
@@ -15,10 +17,13 @@ def connect_elasticsearch():
...
+        if _es.ping():
+            logger.info("Successfully connected to Elasticsearch")
+        else:
+            raise ServiceUnreachable("Elasticsearch")
+    except ServiceUnreachable as e:
+        log_error(e)
```
longshuicy, line 26: i think you are missing an `await` here. log_error function is async. I set an break point here and it didn't actually enter this function :-(

max-zilla: thank you, i added the await. we can talk today about how to test this

PR #181, opened 2022-11-03
175 search box syntax

longshuicy:
- Toggle between Lucene query string and normal facet search
- Move error boundary to only wrap around the results part
- When string incomplete showing some hints and error message from elasticsearch
- 
image
image
image


ddey2:
I am reviewing your PR. SOme things I noted are:

1. I searched for 'dataset' in searchbox and the same results appear multiple times. I checked the elasticsearch index and there are only 4 entries for datasets.  One reason could be, they had been updated and the search result show all of the combinations. Do we want that?

Screen Shot 2022-11-03 at 11 56 11 AM

2. Not able to see 'Author' field in facet
Screen Shot 2022-11-03 at 11 59 45 AM

3. I typed 'author:ddye2' and it showed "No suggestions". It gave me results upon hitting 'ENTER'. Also, I just uploaded a file, but can't seem to find it in results. I thought we are searching both file and datasets.
Screen Shot 2022-11-03 at 12 02 59 PM

longshuicy:
> 2\. Not able to see 'Author' field in facet

Luigi also had the same issue. Did you try clear all of your elasticsearch database and start from scratch? If yes, I will open an issue to investigate this

longshuicy:
> 3\. I typed 'author:ddye2' and it showed "No suggestions". It gave me results upon hitting 'ENTER'. Also, I just uploaded a file, but can't seem to find it in results. I thought we are searching both file and datasets.

Yes we are searching both file and datasets.
1. Looking at screenshot you are not switching to the query string search? Facet search probably didn't understand "author:ddye2" this syntax. 

2. for file it's ` "creator": {"type": "keyword"},` according to the config. Author is only for dataset. We might want to think about unify those field name.

longshuicy:
> 1. I searched for 'dataset' in searchbox and the same results appear multiple times. I checked the elasticsearch index and there are only 4 entries for datasets.  One reason could be, they had been updated and the search result show all of the combinations. Do we want that?

I also noticed that... it has weird <Mark/> related to that too... i opened an issue to investigate that further https://github.com/clowder-framework/clowder2/issues/185

ddey2:
Thanks for clarification on no3. I will clear my cache and again check for no.2

lmarini (approved):

PR #184, opened 2022-11-03
170 add extractor to gui

tcnichol:


longshuicy:
@tcnichol I fixed a few things for you. you can look at the git history to see the diff. 
Mostly it's you didn't connect the listener correctly to the reducer. Everything else looks good.

Now you look at the network tab you should see it's talking to your listen endpoint

image

frontend/src/actions/extractors.js
```diff
@@ -0,0 +1,24 @@
...
+import {handleErrors} from "./common";
+import config from "../app.config";
+import {getHeader} from "../utils/common";
+
+export const RECEIVE_EXTRACTORS = "RECEIVE_EXTRACTORS";
+
+export function fetchExtractors(){
+	return (dispatch) => {
+		// TODO: Parameters for dates? paging?
+		return V2.ExtractorsService.getExtractorsApiV2ExtractorsGet()
```
longshuicy, line 11: There is no method called `getExtractorsApiV2ExtractorsGet` exist. Could you double check the backend route. Is there any GET endpoint exist?

tcnichol: I think this one went away and what I have is from before. i'll just remove this one since we won't be using it anywhere.

tcnichol:
In order to test this, you will need to use the pyclowder from this pull request:
[](https://github.com/clowder-framework/pyclowder/pull/51
)
to register extractors, run the /app/rabbitmq/heartbeat_listener_sync.py and then run an extractor. That will register the extractor.

Files and datasets can be submitted.

tcnichol:
Instructions on how to test.

This pull request depends on this pull request in pyclowder. 

Without using that pyclowder, the extractors will not work with clowder v2.

To register an extractor:

create a new configuration to run the script: 

 /app/rabbitmq/heartbeat_listener_sync.py

I had issues with imports from clowder2, but if I opened it up in /backend in PyCharm I was able to run it without issues. 

While that is running, start an extractor (example - wordcount). You will then see that it is registered in the console.

After that point, you don't need to run the heartbeat monitor. 

If you run clowder, you can see the extractors/listeners under the page:

localhost:3000/listeners

You can submit a file or dataset to an extractor. There is a button for each of them for EXTRACT.

Right now the GUI is pretty bad, and it's not possible to support parameters through the GUI even though the backend routes support them. 

Also, every now and then pyclowder will throw an error 'host is empty this is bad.' Restarting everything should fix that. I will make that a separate issue.

Additionally, right now the metadata for the extractor is not visible under the 'metadata' tab. Looking for input as to how to display it and where.

longshuicy (approved):
can we merge this as it is and work on top of that?
basic things works

lmarini:
> can we merge this as it is and work on top of that?

@longshuicy go for it, please squash

max-zilla (approved):

PR #187, opened 2022-11-04
fix frontend build

longshuicy:
cherry-picked from helm branch


frontend/src/app.config.ts
```diff
@@ -13,7 +13,7 @@ interface Config{
 }
 
 const config:Config = <Config>{};
-const hostname = process.env.CLOWDER_REMOTE_HOSTNAME || "";
+const hostname = process.env.CLOWDER_REMOTE_HOSTNAME || `${window.location.protocol}//${window.location.host}`;
```
longshuicy, line 16: after another thought, maybe one file is enough. How about this?

robkooper: looks good. lets try it

robkooper (approved):

PR #213, opened 2022-11-17
206 modify time facet not working properly

longshuicy:
image


ddey2:
@longshuicy I looked into this. It's working properly. I was wondering if we can put 'Created time' somewhere, to give some text to user what dates we are using. Else looks good to me.

lmarini (approved):

PR #214, opened 2022-11-17
202 file breadcrumb folders not in breadcrumb for files fixed

tcnichol:
This is fixed. The file page was getting folderpath from dataset.state instead of folder.state.


longshuicy (approved):

lmarini (approved):

tcnichol:
Checked. The breadcrumb is right after reload now. Also checked with subfolder.

However, navigation seems broken after you upload file to folder and it redirects.

longshuicy (approved):

PR #231, opened 2022-12-03
230 add public or private option to dataset

tcnichol:
I am opening this as a draft pull request. My plan is to limit the scope of this particular issue to being able to make datasets public (visible to everyone) or private (visible to user only),  giving users the ability to change status and having the public ones show and the private ones hidden. Giving users other than the creator the ability to do anything but view a public dataset will be saved for later. 

Right now I need to replace the text input for status with a dropdown. Will look for feedback,.


ddey2:
@tcnichol @lmarini Is this ready for review? I see it's in draft state

tcnichol:
I need to merge in main and make a few small fixes, then this should be ready.

tcnichol:
I'm marking this one ready for review.

You can now select between status of PUBLIC or PRIVATE with a SHARED option currently commented out. I wanted to make sure this was a dropdown and not a text input. Current behavior is that if a dataset is PUBLIC anyone can see it, and PRIVATE only the creator can share it. Other options would need to wait for roles to be enabled to work. 

Visually it's not that great, so hoping to get some feedback to make it look better and more consistent with the rest of the front end design.

lmarini (changes requested):
Everything works great with the exception of changing status to the default value. If I don't manually select PRIVATE in the pull down, it doesn't switch to it.

tcnichol:
Luigi, I was unable to replicate that error.

Will also be putting a check in place where (for now) only the owner can change something from private to public or back again. This will probably need to be changed once permissions are implemented, but would make sense for now.

lmarini:
@tcnichol not sure why, but I am also not able to replicate that exact error. Sorry for the confusion, I must have not had the latest version. 

I think I am seeing something related. If I switch one dataset and that dataset is set to public, I think that's the default if I go to switch another dataset, instead of the current state of that dataset.

lmarini:
Do you still want to add the check so that only owner can switch state to this PR?

tcnichol:
Luigi,

Right now only the owner being able to change status is enforced on the backend. I think it would be ideal if the option doesn't even appear on t

PR #239, opened 2022-12-12
added new fields to metadata

ddey2:
1. Added new fields to metadata in elasticsearch
2. Updated fields in metadata index in elasticsearch on updating file/dataset in Clowder GUI

### DIFF
--- a/backend/app/routers/datasets.py
+++ b/backend/app/routers/datasets.py
@@ -32,7 +32,6 @@ from rocrate.rocrate import ROCrate
 from app import dependencies
 from app import keycloak_auth
 from app.search.connect import (
-    connect_elasticsearch,
     insert_record,
     delete_document_by_id,
     update_record,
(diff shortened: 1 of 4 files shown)


ddey2:
will address the backend lint failure on next commit along with other feedbacks

tcnichol:
> will address the backend lint failure on next commit along with other feedbacks

I went ahead and ran pipenv run black app and pushed since it was a simple fix. Should finish review on this soon.

ddey2:
Thanks @tcnichol

tcnichol (approved):
This one works and all tests are passing so marking approved.

longshuicy (approved):

backend/app/routers/datasets.py
```diff
@@ -372,13 +383,27 @@ async def patch_dataset(
...
-                    "description": ds["description"],
+                    "name": dataset["name"],
+                    "description": dataset["description"],
                     "author": UserOut(**user).email,
-                    "modified": ds["modified"],
+                    "modified": dataset["modified"],
                 }
             }
             update_record(es, "dataset", doc, dataset_id)
+            # updating metadata in elasticsearch
```
lmarini, line 393: I think this code block is repeated a couple of times in a couple of files. Would it make sense to create a function in a common `util.py` module?

lmarini (changes requested):
Just a small suggestion about duplicate code.

backend/app/routers/datasets.py
ddey2: Yes, the exact code is repeated twice in datasets.py

I was thinking about creating a common function but then realized the doc would be different everywhere, hence thought of keeping it like this. If you think it would be needed at more places, I can create an utility function.

lmarini (approved):

PR #240, opened 2022-12-13
deleting metadata on file/dataset deletion

ddey2:


tcnichol (approved):
Tested and everything seems to be working. Marking approved.

backend/app/search/connect.py
```diff
@@ -124,6 +124,19 @@ def delete_document_by_id(es_client, index_name, id):
...
+    Arguments:
+        es_client -- elasticsearch client which you get as return object from connect_elasticsearch()
+        index_name -- name of index you want to delete
+        query -- query to be searched
+    """
+    try:
+        es_client.delete_by_query(index=index_name, query=query)
+    except BadRequestError as ex:
+        logger.error(str(ex))
+
```
longshuicy, line 138: Do you want to catch more errors other than badrequestError?
Maybe add another `except: logger.error(str(e))`?

Also not throwing this error?

longshuicy (approved):

backend/app/search/connect.py
lmarini: @ddey2 was this resolved?

ddey2: @lmarini Not resolved. @longshuicy We don't perform the action if the service is unavailable. In this scenario, I think BadRequestError is sufficient. I am all ears to suggestion. Lmk if you think I am missing any scenario

Do we want to throw the exception? We show a message if the Elasticsearch service is not available.

lmarini (approved):

PR #245, opened 2022-12-15
241-error-on-delete-metadata fixed.

tcnichol:
I noticed that an error was thrown when deleting metadata. Metadata_In didn't have fields for extractor or extractor_info, adding those has fixed the error. Metadata can be deleted from dataset or file


ddey2:
@tcnichol I am reviewing this. Do I need to add any specific metadata for this to be tested?

tcnichol:
No. The error came up on any metadata deletion, so any metadata would be fine.

lmarini (approved):

PR #247, opened 2022-12-15
Fix/docker compose

lmarini:
Fixed issues with running prod with docker compose.

- fixed elasticsearch URL
- use separate keycloak postgres database for prod and dev. This might require deleting existing volumes `docker compose down -v`
- restart all services if error so that backend can eventually connect to elasticsearch container (which takes a long time to come online)

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -68,7 +68,7 @@ services:
       volumes:
         - postgres_data:/var/lib/postgresql/data
       environment:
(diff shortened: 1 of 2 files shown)


longshuicy:
i couldn't make it work. Got these error:
image

lmarini:
> 

@longshuicy backend will be unable to connect to elasticsearch until elasticsearch comes up. Backend should restart until it can connect.

ddey2 (approved):

PR #249, opened 2022-12-15
fix to deleting metadata in elasticsearch

ddey2:


tcnichol:
An error comes up when deleting metadata but there is another pull request for that bug:f
https://github.com/clowder-framework/clowder2/pull/245

With those fixes this pull request works and properly deletes from elasticsearch. Should I mark this one approved, or wait until that one is merged into main and then main is merged in here?

ddey2:
@tcnichol I reviewed and approved the #245 
I believe merging both should be good.

lmarini (approved):

PR #258, opened 2022-12-20
Initial roles and dependencies

lmarini:
Simple dataset/user/role model and fastapi dependencies to protect backend endpoints. Enforces implied role hierarchy OWNER > EDITOR > UPLOADER > VIEWER.

Includes new `/authorizations` router, new model, and two new dependencies to require a certain role on an endpoint. Examples or its use are in `routers.authorization.py` and `test_authorization.py`.


tcnichol:
I notice that when i create files or datasets, the author is listed as 'undefined undefined'. In the db, the users don't have first name or last name. Will debug this and see if I can find the cause.
undefined_name

ddey2 (changes requested):

backend/app/models/users.py
```diff
@@ -11,11 +11,11 @@
 
 class UserBase(MongoModel):
     email: EmailStr
-    first_name: str
```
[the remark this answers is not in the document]
lmarini: It was requiring to provide first and last name when login in using the `/login` endpoint. We can add the same fields to UserOut. Thanks for pointing this out.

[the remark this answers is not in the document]
lmarini: Should be fixed.

lmarini:
> I notice that when i create files or datasets, the author is listed as 'undefined undefined'. In the db, the users don't have first name or last name. Will debug this and see if I can find the cause. undefined_name

Should be fixed. As mentioned above "It was requiring to provide first and last name when login in using the /login endpoint. We can add the same fields to UserOut. Thanks for pointing this out."

ddey2 (approved):
Looks good

tcnichol:
I checked out the latest code. When I register a new user it still ends up without first_name or last_name in the users db table, and then I get errors when trying to create a dataset. 

``
  File "/Users/helium/.virtualenvs/backend-qaHHe5b_/lib/python3.9/site-packages/fastapi/dependencies/utils.py", line 543, in solve_dependencies
    solved = await call(**sub_values)
  File "/Users/helium/ncsa/clowder2/backend/./app/keycloak_auth.py", line 94, in get_current_user
    user = await get_user_out(userinfo["email"], db)
  File "/Users/helium/ncsa/clowder2/backend/./app/models/users.py", line 43, in get_user_out
    return UserOut(**user_out)
  File "/Users/helium/ncsa/clowder2/backend/./app/models/mongomodel.py", line 30, in __init__
    super().__init__(**pydict)
  File "pydantic/main.py", line 342, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 2 validation errors for UserOut
first_name
  field required (type=value_error.missing)
last_name
  field required (type=value_error.missing)
``

tcnichol:
@lmarini 

I fixed one of the tests. There was a missing import from elasticsearch.

lmarini:
All test are working now. Thanks @tcnichol .

tcnichol:
Luigi, I notice that here it shows all tests passed, but locally I am still getting many tests failing with this error:

``
 TypeError: app.models.users.UserOut() argument after ** must be a mapping, not NoneType

``

tcnichol (approved):
After deleting local volumes, all tests passed locally. I am marking this approved. New users register properly.

backend/app/deps/authorization_deps.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+class Authorization:
+    """We use class dependency so that we can provide the `permission` parameter to the dependency.
+    For more info see https://fastapi.tiangolo.com/advanced/advanced-dependencies/."""
+
+    def __init__(self, role: str):
+        self.role = role
+
+    async def __call__(
```
longshuicy, line 29: nice use of the `__call__`!

backend/app/deps/authorization_deps.py
```diff
@@ -0,0 +1,66 @@
...
+        self,
+        dataset_id: str,
+        db: MongoClient = Depends(get_db),
+        current_user: str = Depends(get_current_username),
+    ):
+        authorization = await db["authorization"].find_one(
+            {"dataset_id": dataset_id, "user_id": current_user, "creator": current_user}
```
longshuicy, line 36: Do we only consider dataset for now?

lmarini: Yes. As mentioned in the design discussions, users will only be able to give access to datasets and get access to all folders/data/metadata. Eventually we can try to extend this. But we should try to get it right for this use case first.

backend/app/routers/authorization.py
```diff
@@ -0,0 +1,55 @@
...
+
+@router.post("", response_model=AuthorizationDB)
+async def save_authorization(
+    authorization_in: AuthorizationBase,
+    user=Depends(keycloak_auth.get_current_username),
+    db: MongoClient = Depends(dependencies.get_db),
+):
+    """Save authorization info in Mongo. This is a triple of dataset_id/user_id/role."""
+    authorization_db = AuthorizationDB(**authorization_in.dict(), creator=user)
+    new_authorization = await db["authorization"].insert_one(
```
longshuicy, line 23: Should we check if user exist first before insert?

longshuicy (approved):
All the endpoints works so approving. 
Left a few comments

backend/app/routers/authorization.py
lmarini: Not sure. `get_current_username` gets the user id from the token, so in theory it should exist, since the token was created. That method throws an exceptions if there is a KeycloakAuthenticationError. We could also use `get_current_user` which goes to Mongo, but I am not sure that is necessary.

backend/app/deps/authorization_deps.py
lmarini: 👍

max-zilla (approved):
I like the Provenance shared model implementation.

PR #264, opened 2023-01-05
253 implement extractor filtering by category

longshuicy:
This PR does the following:
- backend on the endpoint `GET /listeners` add new query parameter `category` to enable filtering
- backend create new endpoint `GET /listeners/category` to list all the distinct categories in the whole listener collection
- backend create new endpoint `GET /listeners/defaultLabels` to list all the distinct labels in the whole listener collection
- frontend implement the filter radio button groups
- `search` and `filter` are independent of each other: i.e. when searching filter will jump to "all"; when filtering the searchbox will clear out

I noticed that there is only "EXTRACT" as the category in most of the V1 extractors; mannually added geospatial as an example

image


longshuicy:
This is based off PR 252; just for ease of review i point this to 252

lmarini (approved):

PR #274, opened 2023-01-10
260 show extractor metadata

tcnichol:
With this pull request, metadata from extractors/listeners is visible under a new tab. Right now the data can be viewed, with each entry having a toggle to  view/hide metadata. 

For metadata, the following file includes some sample extractor metadata. Some of the fields are not meaningful but were included to make the json nested.
[file_metadata.txt](https://github.com/clowder-framework/clowder2/files/10385817/file_metadata.txt)

### DIFF
--- a/.run/uvicorn.run.xml
+++ b/.run/uvicorn.run.xml
@@ -6,9 +6,9 @@
     <envs>
       <env name="PYTHONUNBUFFERED" value="1" />
     </envs>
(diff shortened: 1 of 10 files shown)


tcnichol:
I realize I need to add the tab for datasets - right now it just works for files. Should be done soon.

tcnichol:
Extractor metadata added for datasets now.

lmarini (approved):

PR #279, opened 2023-01-14
276 automatically trigger extractor whenever a file is updated

tcnichol:
With this pull request, if a file is updated then all extractors that were run on this file will be ran again. 

Right now we are not saving information on the parameters used if an extractor has parameters, so for this pull request the resubmissions won't include parameters. Adding those will be dependent on other issues currently in progress. 

In order for this to work, a listener/extractor needs to be in the db. Here is a sample of one (wordcount) that can be used. Add to the 'listeners' db table.

``
{
  "_id": {
    "$oid": "63b5cd4aeb1180d52266214e"
  },
  "author": "Rob Kooper <kooper@illinois.edu>",
  "name": "ncsa.wordcount",
  "version": "2.0",
  "description": "WordCount extractor. Counts the number of characters, words and lines in the text file that was uploaded.",
  "creator": null,
  "created": {
    "$date": {
      "$numberLong": "1672858954451"
    }
  },
  "modified": {
    "$date": {
      "$numberLong": "1672858954451"
    }
  },
  "properties": {
    "author": "Rob Kooper <kooper@illinois.edu>",
    "process": {
      "file": [
        "text/*",
        "application/json"
      ]
    },
    "maturity": "Development",
    "name": "ncsa.wordcount",
    "contributors": [],
    "contexts": [
      {
        "lines": "http://clowder.ncsa.illinois.edu/metadata/ncsa.wordcount#lines",
        "words": "http://clowder.ncsa.illinois.edu/metadata/ncsa.wordcount#words",
        "characters": "http://clowder.ncsa.illinois.edu/metadata/ncsa.wordcount#characters"
      }
    ],
    "repository": [
      {
        "id": {
          "$oid": "63b5cd4aeb1180d52266214d"
        },
        "repository_type": "git",
        "repository_url": ""
      }
    ],
    "external_services": [],
    "libraries": [],
    "bibtex": [],
    "default_labels": [],
    "categories": [],
    "parameters": {
      "schema": {
        "X_MIN_START": {
          "type": "integer",
          "title": "X_MIN_START"
        },
        "X_MIN_END": {
          "type": "integer",
          "title": "X_MIN_END"
        },
        "Y_MIN_START": {
          "type": "integer",
          "title": "Y_MIN_START"
        },
        "Y_MIN_END": {
          "type": "integer",
          "title": "Y_MIN_END"
        },
        "ZONE": {
          "type": "string",
          "title": "ZONE"
        }
      }
    },
    "version": "2.0"
  }
}
``


lmarini (changes requested):
Please add comments to code. Thank you.

tcnichol:
comments added. In this pull request the extractors are re-run using a helper method. There is also a route for re-running extractors added; it seemed like it might be useful.

Once the Job is implemented, I will be able to use parameters of former extractor runs.

backend/app/routers/files.py
```diff
@@ -345,3 +390,46 @@ async def get_file_extract(
...
+    implemented.
+
+        Arguments:
+        file_idd: Id of file
+        credentials: credentials of logged in user
+        db: MongoDB Client
+        rabbitmq_client: Rabbitmq Client
+
+    """
+    if (file := await db["files"].find_one({"_id": ObjectId(file_id)})) is not None:
```
longshuicy, line 414: This seems to be similar to the private method `_resubmit_file_extractors`. Can you reuse that?

backend/app/routers/files.py
```diff
@@ -207,6 +251,7 @@ async def update_file(
             }
         }
         update_record(es, "file", doc, updated_file.id)
+        await _resubmit_file_extractors(file_id, credentials, db, rabbitmq_client)
```
longshuicy, line 254: It might be worthy to add a return message to the `_resubmit...` function so you can get some status back? If resubmit failed there will be a way to return that info to the client?

tcnichol: I made the method return a list. If any extractor succeeds, it goes into a succeeds list and the ones that fail go into a fail list.

backend/app/routers/files.py
tcnichol: It is re-used now. Unnecessary code also taken out.

backend/app/routers/files.py
```diff
@@ -43,6 +43,57 @@
...
+    implemented.
+
+        Arguments:
+        file_id: Id of file
+        credentials: credentials of logged in user
+        db: MongoDB Client
+        rabbitmq_client: Rabbitmq Client
+
+    """
+    if (file := await db["files"].find_one({"_id": ObjectId(file_id)})) is not None:
```
max-zilla, line 64: In the spot where you call this method, you already have the FileOut you could pass in directly and not have to query it again here.

backend/app/routers/files.py
```diff
@@ -43,6 +43,57 @@
...
+        file_out = FileOut.from_mongo(file)
+        query = {"resource.resource_id": ObjectId(file_id)}
```
max-zilla, line 66: this will match all metadata regardless of resource.version, so any previous file version metadata will also be deleted. is this the intent?

max-zilla (commented):
A couple small comments, most of it looks good otherwise.

backend/app/routers/files.py
tcnichol: fixed that.

backend/app/routers/files.py
tcnichol: I notice that MetadataOut doesn't seem to have the version. I will add that and then check to make sure this is the latest.

tcnichol: Added that, just need to test to make sure it works.

tcnichol:
I merged 'main' into this branch and am noticing that it no longer seems to work correctly. 

Will try to fix tomorrow before the clowder meeting.

max-zilla:
@tcnichol please resolve merge conflict

max-zilla (approved):

Issue #277, opened 2023-01-12
Keep track of old versions of user metadata in mongo 


ddey2:
to be consistent with what we do in extractor metadata.

PR #280, opened 2023-01-17
Metdata version updated with metadata update 

ddey2:
1. Metadata version updated in mogodb with every metadata update
2. Show latest available metadata if metadata with specific file version is not available


tcnichol:
I think doing the black formatting should fix the build error, then I can mark this one approved.

ddey2:
> I think doing the black formatting should fix the build error, then I can mark this one approved.

No idea why it's failing for a file where I haven't made any changes. However, still trying to format the file.

longshuicy:
Sorry for not understanding the context, do you mean that if metadata has been changed, the file version need to be automatically bumped up? If so could you point me to the part that `file_version` has been updated?

ddey2:
> Sorry for not understanding the context, do you mean that if metadata has been changed, the file version need to be automatically bumped up? If so could you point me to the part that `file_version` has been updated?

There is 2 parts to this PR:

1. I enabled the 'all_version' from UI end so that it captures the file version whenever there is a request to update metadata.
2. We want to update the version in 'metadata' table in mongodb, current scenario we weren't updating the version. The reason we want to update it - If a file is updated, we first want to fetch metadata associated with that specific version, if it's not available, we want to show the latest version metadata.

I hope that answers your question. Lmk if you have any more.

frontend/src/openapi/v2/services/MetadataService.ts
```diff
@@ -118,7 +118,7 @@ export class MetadataService {
      */
     public static getFileMetadataApiV2FilesFileIdMetadataGet(
         fileId: string,
-        allVersions: boolean = false,
+        allVersions: boolean = true,
```
ddey2, line 121: @longshuicy I enabled this here to send the version with the request to update metadata

backend/app/routers/metadata_files.py
```diff
@@ -241,6 +241,20 @@ async def update_file_metadata(
     Returns:
         Metadata document that was updated
     """
+
+    # check if metadata with file version exists, replace metadata if none exists
```
ddey2, line 245: replace_metadata is the function which is already taking care of updating resource object etc. 

If there is no specific version of metadata is available in mongodb, we want to update the version, hence called 'replace_metadata'

lmarini:
@ddey2 can you resolve conflicts? I am not seeing the black check. Does it only run it if there is no conflict?

tcnichol:
I added a file and added metadata. I updated the file, and when i changed the metadata it updated the version to v2. Is this the expected behavior? Or should the file v1 metadata still be there somewhere?

ddey2:
> I added a file and added metadata. I updated the file, and when i changed the metadata it updated the version to v2. Is this the expected behavior? Or should the file v1 metadata still be saved in the db?

This should be the expected behavior for now (as per our discussion). We plan to keep all the metadata versions, but that's for future. I have created an issue for that #277

tcnichol (approved):
Tested and this work. Marking approved.

max-zilla (changes requested):

backend/app/routers/metadata_files.py
```diff
@@ -248,6 +248,20 @@ async def update_file_metadata(
...
+    if (
+        version_md := await db["metadata"].find_one(
+            {
+                "_id": ObjectId(metadata_in.metadata_id),
+                "resource.resource_id": ObjectId(file_id),
+                "resource.version": metadata_in.file_version,
```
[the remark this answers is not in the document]
ddey2: It's necessary because I need to find out if metadata with the specific file id and version id exists. If it does, we can simple update the metadata here in this method, else I need to call 'replace_file_metadata' where it creates a new object with the file id  and version. I could have updated the current method but chose to re-use what we already have.

max-zilla (approved):

PR #282, opened 2023-01-18
Add extractor job ids

max-zilla:
WIP

Implements an extraction job class with a unique ID that will be associated with an extraction. Extractor status messages ("10% complete...") will be stored individually with the job ID, but we will also have a higher-level collection of the Jobs that include the ID, latest message + status, times, user, resource, etc.

Details:
- when extraction job is sent to an extractor, a Job record is also created in db with its own UUID.
- this will also check whether Clowder has a unique 10-digit instance ID yet, and generate one into the new Mongo config collection if it doesn't exist. this also creates a persistent queue in RabbitMQ called `clowder.[InstanceID]` which will be sent to the extractors as the reply queue
- api/listeners/instance endpoint that tells this unique ID
- `app/rabbitmq/message_listener_sync.py` is the standalone service similar to heartbeat listener. it will also create the instance ID and reply queue if necessary, then listen for messages from extractors and put them into the `listener_job_updates` collection. also the job summary itself in `listener_jobs` will be updated with latest message and timestamp.


backend/app/rabbitmq/message_listener_sync.py
```diff
@@ -0,0 +1,86 @@
...
+
+    job_id = msg["job_id"]
+    status = msg["status"]
+    timestamp = datetime.strptime(msg["start"], '%Y-%m-%dT%H:%M:%S%z')  # incoming format: '2023-01-20T08:30:27-05:00'
+
+    mongo_client = MongoClient(settings.MONGODB_URL)
+    db = mongo_client[settings.MONGO_DATABASE]
+
+    # Check if the job exists, and update if so
+    existing_job = db["listener_jobs"].find_one({"_id": ObjectId(job_id)})
```
longshuicy, line 33: maybe instead of update it just keep storing them? This way we can have a complete history of the job status?

backend/app/routers/listeners.py
```diff
@@ -23,6 +26,28 @@
...
+    # Check all connection and abort if any one of them is not available
+    if db is None:
+        raise HTTPException(status_code=503, detail="Service not available")
+        return
+
+    if (instance_id := await db["config"].find_one({"key": "instance_id"})) is not None:
+        return ConfigEntryOut.from_mongo(instance_id).value
+    else:
+        # If no ID has been generated for this instance, generate a 10-digit alphanumeric identifier
+        instance_id = ''.join(random.choice(string.ascii_uppercase + string.ascii_lowercase + string.digits) for _ in range(10))
```
longshuicy, line 43: is there a convention for the random id? should we use uuid or anything else? thoughts?

backend/app/rabbitmq/message_listener_sync.py
max-zilla: the individual events are also stored in db["listener_job_updates"] but the "listener_jobs" just contain the "header" of each job, with the latest message/status, so we have both if I understand correctly.
```
# Add latest message to the job updates
        event_msg = EventListenerJobUpdate(job_id=job_id, status=status, timestamp=timestamp)
        db["listener_job_updates"].insert_one(event_msg.to_mongo())```

File: backend/app/routers/listeners.py (hunk 2 above)
Comment (max-zilla) [reply backend/app/routers/listeners.py:43]: not sure... this particular snippet is same logic Clowder 1 was using.

File: backend/app/rabbitmq/listeners.py [hunk 3]
```diff
@@ -1,77 +1,151 @@
...
     # TODO check if extractor is registered
-    current_filename = file_out.name
-    current_fileSize = file_out.bytes
+
+    # Create an entry in job history with unique ID
+    job = EventListenerJob(
+        listener_id=routing_key,
+        creator=user,
+        resource_ref=MongoDBRef(
+            collection="file", resource_id=file_out.id, version=file_out.version_num
```
longshuicy, line 74: For file, the frontend will need more information to be able to link to that file page (namely datasetId and folderId)

max-zilla: We will add a specific FileDBRef to do this.

longshuicy (approved):

longshuicy:
Approved since i'm already using some of the models here

PR #286, opened 2023-01-23
284 rabbitmq client configs should come from settings

### LINKED ISSUE #284: Rabbitmq client configs should come from settings
Currently rabbitmq client is hard coded to localhost/guest/guest. We do have fields defined in settings.

We also want to make sure the message sent to rabbitmq includes the correct host. This is required for running backend on host and extractor in container.

lmarini:
Use config variables to setup RabbitMQ client. Were being hardcoded to localhost/guest/guest. Closes #284.


longshuicy (approved):
looks good!

max-zilla (approved):

PR #291, opened 2023-01-24
281 css style change when using rjsf

longshuicy:
Overwrite the default widget of RJSF library using more clowder like components (so far only textbox and select; but can write more for the future)

This also fix the background style bug when changing things on the extractor submission form 

image
image
image


tcnichol (approved):
Tested this one and everything looks good.

PR #294, opened 2023-01-26
271 extractor catalog page on frontend with extractor history

longshuicy:
This PR **add one more endpoint to get all the job Ids** with a few filters
Frontend-wise it **shows a page list all the past executions of the listeners**
Use `date-fns` library to parse the date; and allow user to filter by dates

TODOs:
~~- the filter by submitted date doesn't work yet because i need to correctly parse the frontend date and do a range comparison in the endpoint.~~
- need more information about resources (if it's file, we need the datasetID as well as the folderID for the frontend to link to that file) We might want to discuss this add address it in the next task
- 
image
image


max-zilla (approved):
looks great! date/status filters also worked.

PR #305, opened 2023-02-02
Add mime type feed support

max-zilla:
This will generate a feed from v1 extractors with extractor_info automatically. 

An example is wordcount extractor, which has this in extractor info:
```
"process" : { "file" : [ 
  "text/*", 
  "application/json" 
] }
```

This is translated into the following feed when the heartbeat is first registered:
```
{ "_id" : ObjectId("63dbce508120dca02abc52a2"), 
"name" : "ncsa.wordcount", 
"search" : { 
  "index_name" : "file", 
  "criteria" : [
     { "field" : "content_type_main", "operator" : "==", "value" : "text" }, 
    { "field" : "content_type", "operator" : "==", "value" : "application/json" }
   ], 
  "mode" : "or", 
  "original" : null }, 
"listeners" : [ { "listener_id" : ObjectId("63dbce508120dca02abc52a0"), "automatic" : true } ], 
"author" : null, 
"updated" : ISODate("2023-02-02T14:53:04.282Z") }
```


longshuicy (approved):
I just read the code but haven't tested it. Looks good to me.

ddey2 (commented):
I tested it. It was able to register the extractor in db. Need to fix the logger issue while running heartbeat_listerner_sync.py

ddey2 (approved):

PR #310, opened 2023-02-04
Fix to file deletion

ddey2:
Passed elasticsearch client as parameter while calling remove_file_entry


tcnichol (approved):
Tested and files delete properly now. marking approved.

longshuicy (approved):

PR #322, opened 2023-02-10
update file content model

longshuicy:
Now the content type is not a simple string instead an object that has two fields. 
Update all the codegen models as well to reflect that.

To test:

Just open any datasets that has file in it.


tcnichol:
@longshuicy I checked this one and it fixes the bug. Could someone review this and get it merged in?

tcnichol:
Ran into an issue:

When I try to submit a file to an extractor, I get the blank page again. Will get a bit and post more comments.

tcnichol:
@longshuicy 
I was testing so I pushed 2 small changes. A method somehow got removed from the FilesService for extraction, and posting metadata also caused the blank page.

longshuicy:
> @longshuicy I was testing so I pushed 2 small changes. A method somehow got removed from the FilesService for extraction, and posting metadata also caused the blank page.

Looks good to me. I can't approve cuz it's my branch :-p. But you can approve

max-zilla (approved):

PR #326, opened 2023-02-15
download older file versions & metadata

max-zilla:
No longer allow changing of filename or MIME type when file version changes.

Adds version parameter to file download endpoint. 

Some fixes to heartbeat container for docker-compose.dev.


ddey2:
@max-zilla Your code looks good to me for the issues addressed in the description. I am testing it. 

One question - Can somebody download the older file version metadata from this as suggested in the title? I didn't see the respective code for that.

max-zilla:
> @max-zilla Your code looks good to me for the issues addressed in the description. I am testing it.
> 
> One question - Can somebody download the older file version metadata from this as suggested in the title? I didn't see the respective code for that.

the API already had support for that, it turns out. i don't have a frontend implementation yet.

longshuicy (approved):

ddey2 (approved):

PR #328, opened 2023-02-15
Context matches v1 type.

tcnichol:
In clowder v1 context was a list that consisted in URL and dictionary. This pull request makes v2 match v1. Front end and back end have both been changed. Extractor registration should work correctly.

If testing, make sure to remove old metadata/listeners from DB, or else conflicts will happen.


longshuicy:
Can you fix the linting?

tcnichol:
To test with extractors, use this branch of pyclowder:

https://github.com/clowder-framework/pyclowder/pull/60

Wordcount and other sample extractors there can be used.

tcnichol:
For the errors in pytest. These are happening when metadata with no context_url is being posted to the backend for files and datasets. The field is an optional[str] but the validator looks like it throws an error if it is missing. Not sure what to do to fix it. Also that the changes made in this PR shouldn't affect that.

tcnichol:
@ddey2 I pushed a small change that might fix the errors you are seeing.

ddey2 (approved):
Looks good, tested.

frontend/src/components/metadata/CreateMetadataDefinition.tsx
```diff
@@ -118,10 +118,12 @@ export const CreateMetadataDefinition = (): JSX.Element => {
             contextJson[item["term"]] = item["iri"]
         })
 
+		console.log([JSON.stringify(contextJson)]);
+
```
max-zilla, line 122: please remove

max-zilla (changes requested):
everything seemed to work, remove the log and i will merge after tests pass thanks

frontend/src/components/metadata/CreateMetadataDefinition.tsx
tcnichol: removed

PR #329, opened 2023-02-16
Test for v1 extractor triggering

max-zilla:
Add a couple endpoints to feeds router, add a test for full cycle of triggering a job on a text file based on a v1 extractor.


ddey2 (changes requested):
Pytest failing
your new test case in test_extractor is failing due to the below

await db["config"].insert_one(config_entry.to_mongo())
E           TypeError: object InsertOneResult can't be used in 'await' expression

ddey2 (approved):

PR #331, opened 2023-02-16
311 add spinner and disable the upload button when upload files

arunapa:
Screen Shot 2023-02-16 at 10 02 35 AM


longshuicy:
We may need the same thing for "update" the file. This could be another task next. Thanks!

longshuicy (approved):

PR #334, opened 2023-02-16
Adding download button on UI to download correct verison file

ddey2:


frontend/src/openapi/v2/models/FileVersion.ts
```diff
@@ -7,6 +7,7 @@ import type { UserOut } from './UserOut';
 export type FileVersion = {
     id?: string;
     file_id: string;
+	file_name: string;
```
ddey2, line 10: I wanted to download the file by it's file name. Hence, I added the field here but looks like it's not getting populated. @longshuicy Please take a look

ddey2:
@longshuicy Also take a look at the download button. I am not an expert in UI, it may not look so fancy. :D

frontend/src/openapi/v2/models/FileVersion.ts
longshuicy: You shouldn't directly modify anything in the openapi because that's auto generated by the backend. 
If you would need that as an input, you need to modify the backend logic and then reran `npm run codegen:v2:dev`

longshuicy: Ah i see what you are doing. You just need the filename to be passed in to the function? So probably no need to change the backend design. You can see what I did to make it work in this Diff: https://github.com/clowder-framework/clowder2/pull/334/commits/5afb3ccdd2a458cb083aff5a29ba6e8f0450e9d6

longshuicy (approved):

frontend/src/openapi/v2/models/FileVersion.ts
ddey2: That was a another way of passing that filename. Thanks for showing that. :)

PR #341, opened 2023-02-21
adding restart unless stopped for this container

tcnichol:
Sometimes the container for the heartbeat listener would stop and need to be restarted. This should fix the problem.


tcnichol:
I'm seeing some errors in pytest related to keycloak. Not sure what is going on there, looking into it.

longshuicy (approved):

PR #352, opened 2023-02-23
Added new component for file actions

arunapa:
* Tested file download and file delete from File page

* Verified after deleting file, page navigates back to main dataset page

Screenshot of UI:
Screen Shot 2023-02-23 at 12 08 01 PM

### DIFF
--- a/frontend/src/components/files/UpdateFile.tsx
+++ b/frontend/src/components/files/UpdateFile.tsx
@@ -20,7 +20,7 @@ type UpdateFileProps ={
 export const UpdateFile: React.FC<UpdateFileProps> = (props: UpdateFileProps) => {
 	const dispatch = useDispatch();
 	const updateFile = async (formData: FormData, fileId: string|undefined) => dispatch(fileUpdated(formData, fileId));
-    const listFileVersions = (fileId: string | undefined) => dispatch(fetchFileVersions(fileId));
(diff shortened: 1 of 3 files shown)


longshuicy (changes requested):
I noticed that after you update the version histories are not updated until you refresh
image

I think you need to call list versions one more time after you update. Add this in the `UpdateFile.tsx`

```
	const listFileVersions = (fileId: string | undefined) => dispatch(fetchFileVersions(fileId));

        const onSave = async (formData:FormData) => {
                   ...
                   updateFile(formData, fileId);
                   ...
````

File: frontend/src/components/files/UpdateFile.tsx [hunk 1]
```diff
@@ -19,17 +19,19 @@ type UpdateFileProps ={
...
+    const listFileVersions = (fileId: string | undefined) => dispatch(fetchFileVersions(fileId));
 
 	const {fileId, setOpen,} = props;
 
 	const [loading, setLoading] = useState<boolean>(false);
 
 	const onSave = async (formData:FormData) => {
 		setLoading(true);
-		updateFile(formData, fileId);
+		await updateFile(formData, fileId);
```
longshuicy, line 31: Good finding!

longshuicy (approved):

PR #354, opened 2023-02-23
Added group_id to authorization

ddey2:
1. Added list of 'group_ids' (ObjectIds) to Authorization collection in Mongo
2. Each dataset has one entry in Authorization collection with a list of group_ids and a list of users.
3. Dataset_id is now set as Object Id
4. Associated changes to retrieve role etc in routers/authorization.py
5. Added checks to remove duplicates in list in routers/groups and routers/authorization


ddey2:
@max-zilla As mentioned in slack, we decided to go with this implementation for now. It's subject to change in future based on requirements. Keeping this in mind, could you review the PR? I think this will change a couple of the things like your PR #357

backend/app/routers/authorization.py
```diff
@@ -18,23 +23,62 @@ async def save_authorization(
     user=Depends(keycloak_auth.get_current_username),
     db: MongoClient = Depends(dependencies.get_db),
 ):
-    """Save authorization info in Mongo. This is a triple of dataset_id/user_id/role."""
-    authorization_db = AuthorizationDB(**authorization_in.dict(), creator=user)
+    """Save authorization info in Mongo. This is a triple of dataset_id/user_id/role/group_id."""
+
+    # Check all connection and abort if any one of them is not available
+    if db is None:
```
lmarini, line 29: We should start checking this in the dependencies themselves. Otherwise we have to add this code to every endpoint.

backend/app/routers/authorization.py
```diff
@@ -18,23 +23,62 @@ async def save_authorization(
...
+        raise HTTPException(status_code=503, detail="Service not available")
+        return
+
+    # Retrieve users from groups in mongo
+    user_ids: List[EmailStr] = []
+    group_q_list = db["groups"].find({"_id": {"$in": authorization_in.group_ids}})
```
lmarini, line 35: Do we have a mongo index on this field? Given this query we probably need one.

backend/app/routers/authorization.py
```diff
@@ -18,23 +23,62 @@ async def save_authorization(
...
+        async for group_q in group_q_list:
+            group = GroupOut.from_mongo(group_q)
+            user_ids = list(map(lambda user: user.user.email, group.userList))
+    else:
+        raise HTTPException(
+            status_code=404, detail=f"Group {authorization_in.group_ids} not found"
+        )
+
+    authorization_dict = authorization_in.dict()
+    authorization_dict["user_ids"] = list(set(user_ids))
```
lmarini, line 46: Is this section adding the users from each group to the array of users in authorization document? If so, why do we need to do this?

backend/app/routers/authorization.py
```diff
@@ -18,23 +23,62 @@ async def save_authorization(
...
+
+    # Get group id and the associated userList from authorization
+    if (
+        authorization_q := await db["authorization"].find_one(
+            {"dataset_id": ObjectId(dataset_id)}
+        )
+    ) is not None:
+        authorization = AuthorizationDB.from_mongo(authorization_q)
+
+        if current_user in authorization.user_ids:
```
lmarini, line 71: Here we could check the groups, by retrieving the list of groups a user is in and checking if any of those groups are in the Authorization group list.

lmarini (changes requested):
Left some small comments and one bigger one regarding the interaction of list of users and list of groups and how we query them.

backend/app/routers/authorization.py
ddey2: Will create an index on group_ids

backend/app/routers/authorization.py
ddey2: will move this.

backend/app/routers/authorization.py
ddey2: I added list(set(user_ids) so that if we add same user_ids twice, I want to remove the duplicate user_id

backend/app/routers/authorization.py
ddey2: That works too but I am wondering if the way I have implemented it now will not work.

max-zilla (approved):

backend/app/models/authorization.py
```diff
@@ -20,9 +22,10 @@ class RoleType(str, Enum):
 class AuthorizationBase(BaseModel):
     # TODO: This should be PyObjectId = Field(default_factory=PyObjectId). Need to figure out why can't create instance
     #  in `routers.authorization.get_dataset_role()`.
-    dataset_id: str
-    user_id: EmailStr
+    dataset_id: PyObjectId
+    user_ids: List[EmailStr] = []
```
longshuicy, line 26: Ah... we are doing a list of user_ids now? this might break the view :-( cuz I was assuming this is just one string

longshuicy: If we have decided to make user_ids a list, then I'll need to update some of the logic to filter user access. Will convert this into an issue.

longshuicy (approved):

PR #355, opened 2023-02-27
343 list resources that user has access

longshuicy:
I added authorization filter to below endpoints:

get all datasets /datasets
get all folders /datasets/{datasetId}/folders
get all files /datasets/{datasetId}/files
get all listener jobs /jobs

Let me know if we'd like to roll this out to all the get endpoints?

### DIFF
--- a/backend/app/routers/datasets.py
+++ b/backend/app/routers/datasets.py
@@ -278,7 +278,7 @@ async def get_dataset_files(
     files = []
     if folder_id is not None:
         for f in (
-            await db["files_views"]
+            await db["files_view"]
             .find(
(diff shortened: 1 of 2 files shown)


tcnichol:
I have now tested this for datasets, files, folders etc. There are some tests failing but once those are fixed I can mark approved here.

longshuicy:
Can we just merge this since it's pointing to the other "View" PR. In the View PR im attempting to fix the failed pytest

lmarini:
sounds good @longshuicy

lmarini (approved):

PR #356, opened 2023-02-27
Properly interpret extractor statuses

max-zilla:
This will correctly parse pyclowder's status message updates to update job status and duration.

It also moves the message listener into docker-compose so it will start with everything else just like heartbeat listener.


lmarini:
Should we rename `message_listener_sync.py` to `status_listener_sync.py` or something more specific along those lines?

longshuicy:
image
This is working~~ exciting!

May I fix the millisecond issue (as well as find more icons for the status since i noticed we have more status now)?

tcnichol (approved):
approved. checked and this works.

longshuicy (approved):

PR #362, opened 2023-03-01
348 metadata value not visible on UI

tcnichol:
This should now be fixed for all types. I created metadata definitions that are text input and date, and added the Unit metadata definition to test a dropdown. 

Part of the problem was that 'contents' was used in place of 'content' in a few places so now it's always content for all metadata type. Also fieldName is an object, so using fieldName.name for the key for content[fieldName.name] to get actual metadata value. Metadata can now be created, value shown is right, and the updates work.


ddey2 (changes requested):
I tried and got Bad request error. No more info on backend. I deleted all the volumes and restarted docker containers, so it shouldn't have nay issue with db as well.


Screenshot 2023-03-02 at 8 50 58 AM

tcnichol:
The error should be fixed now.

longshuicy (approved):

max-zilla (approved):

PR #363, opened 2023-03-01
only get current version's metadata

max-zilla:
Avoid showing previous versions of file metadata if multiple versions of file.


longshuicy (approved):
Looks good!

PR #370, opened 2023-03-02
changing context to list in metadata defs

tcnichol:
When working on the metadata UI issues I noticed these aren't updated to fit context being a list. Thought it should be its own pull request rather than adding it to that one.


longshuicy:
Hmm this just updates the example?

PR #371, opened 2023-03-03
Added new Page Not Found component, updated route to use new component

arunapa:
Completed ~~Note: PR to be updated with code to explicitly redirect users to 404 page when they request a resource that cannot be found~~

* tested by providing an incorrect file url, verified redirects to page not found

Screen Shot 2023-03-03 at 10 56 31 AM


longshuicy:
hm actually i can't trigger this 404. whenever I gave a wrong URL, i get 500 error. E.g.
```
role = AuthorizationDB.from_mongo(authorization).role
AttributeError: 'NoneType' object has no attribute 'role'
```

longshuicy (approved):

longshuicy:
To test:

1. go to a non-exist routes for example `/datasets` or `/files` or `/something random`
2. edit file id in the URL but you have to match digits to make it a valid objectID

PR #385, opened 2023-03-07
Implement FileAuthorization dependency

max-zilla:
Add FileAuth dependency we can use if file_id is available in a route, and implements it on several file routes.


tcnichol:
I checked and this one looks pretty good, I noticed that the backend lint was failing so I just pushed a commit to fix that.

PR #386, opened 2023-03-07
Implement DatasetAuthorization dependency, add to endpoints

tcnichol:
This pull request adds dataset authorization dependency and adds this to the dataset endpoints.

### DIFF
--- a/backend/app/deps/authorization_deps.py
+++ b/backend/app/deps/authorization_deps.py
@@ -5,7 +5,7 @@ from bson import ObjectId
 from app.dependencies import get_db
 from app.keycloak_auth import get_current_username
 from app.models.authorization import RoleType, AuthorizationDB
-from app.models.files import FileOut
+from app.models.datasets import DatasetOut
 
 
 async def get_role(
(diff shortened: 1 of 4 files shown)


backend/app/deps/authorization_deps.py
```diff
@@ -92,6 +92,41 @@ async def __call__(
             )
 
 
+class DatasetAuthorization:
```
max-zilla, line 95: The base Authorization class that Luigi implemented functions as DatasetAuthorization, you should be able to remove this one and use Authorization instead.

backend/app/routers/datasets.py
```diff
@@ -358,6 +360,8 @@ async def edit_dataset(
     db: MongoClient = Depends(dependencies.get_db),
     user_id=Depends(get_user),
     es=Depends(dependencies.get_elasticsearchclient),
+    allow: bool = Depends(DatasetAuthorization("editor")),
```
max-zilla, line 363: in these cases, you can just use Depends(Authorization( as mentioned above.

backend/app/routers/datasets.py
```diff
@@ -498,6 +506,7 @@ async def add_folder(
     folder_in: FolderIn,
     user=Depends(get_current_user),
     db: MongoClient = Depends(dependencies.get_db),
+    allow: bool = Depends(DatasetAuthorization("editor")),
```
max-zilla, line 509: there is also a role between viewer and editor that is "uploader". I think a few of these could use the more relaxed uploader role instead if they involve only additive things - maybe add_folder, save_file, get_dataset_extract?

backend/app/routers/datasets.py
```diff
@@ -557,6 +567,7 @@ async def delete_folder(
     db: MongoClient = Depends(dependencies.get_db),
     fs: Minio = Depends(dependencies.get_fs),
     es: Elasticsearch = Depends(dependencies.get_elasticsearchclient),
+    allow: bool = Depends(DatasetAuthorization("owner")),
```
max-zilla, line 570: i think it is OK to let editors do this also

max-zilla (changes requested):

backend/app/routers/datasets.py
tcnichol: These are fixed. I don't use the allow yet but I am working on that, will have some questions at the meeting tomorrow.

backend/app/routers/datasets.py
tcnichol: Fixed this.

backend/app/deps/authorization_deps.py
tcnichol: Removed.

backend/app/routers/datasets.py
tcnichol: I put uploader for these.

backend/app/routers/datasets.py
```diff
@@ -358,7 +360,10 @@ async def edit_dataset(
     db: MongoClient = Depends(dependencies.get_db),
     user_id=Depends(get_user),
     es=Depends(dependencies.get_elasticsearchclient),
+    allow: bool = Depends(Authorization("editor")),
 ):
+    if not allow:
+        raise HTTPException(status_code=404, detail=f"Dataset {dataset_id} not found")
```
longshuicy, line 366: 404 here intended? not 403?

tcnichol: Which should these be? I can change them all to 403.

max-zilla (approved):

PR #387, opened 2023-03-07
add MetadataAuthorization dependency

tcnichol:
Add MetadataAuth dependency we can use if metadata_id is available in a route, and implements it on several metadata routes.

Also implemented for some dataset metadata routes.

Files will wait until this pull request merged:

https://github.com/clowder-framework/clowder2/pull/385

### DIFF
--- a/backend/app/routers/datasets.py
+++ b/backend/app/routers/datasets.py
@@ -28,7 +28,7 @@ from rocrate.rocrate import ROCrate
 
 from app import dependencies
 from app import keycloak_auth
-from app.deps.authorization_deps import Authorization
+from app.deps.authorization_deps import Authorization, DatasetAuthorization
 from app.config import settings
 from app.keycloak_auth import get_token
 from app.keycloak_auth import get_user, get_current_user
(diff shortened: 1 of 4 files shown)


backend/app/routers/metadata_datasets.py
```diff
@@ -83,6 +84,7 @@ async def add_dataset_metadata(
     user=Depends(get_current_user),
     db: MongoClient = Depends(dependencies.get_db),
     es: Elasticsearch = Depends(dependencies.get_elasticsearchclient),
+    allow: bool = Depends(DatasetAuthorization("editor")),
```
max-zilla, line 87: These should be MetadataAuthorization I think?

backend/app/routers/datasets.py
```diff
@@ -28,7 +28,7 @@
 
 from app import dependencies
 from app import keycloak_auth
-from app.deps.authorization_deps import Authorization
+from app.deps.authorization_deps import Authorization, DatasetAuthorization
```
max-zilla, line 31: this can be removed (covered in other PR)

max-zilla (changes requested):

max-zilla (approved):

PR #388, opened 2023-03-08
Group management

max-zilla:
PR touches a few things:

- add GroupAuthorization dependency that enforces group members/editors/owners
- use Provenance in groups model
- improvements to several auth/groups endpoints (some were not functioning properly)
- new endpoints for setting user/group roles on a dataset
- new endpoints for adding/removing user from a group
- create test utils file with some shared JSON - need to continue to leverage
-

### DIFF
--- a/backend/app/routers/groups.py
+++ b/backend/app/routers/groups.py
@@ -109,6 +109,8 @@ async def add_member(
     db: MongoClient = Depends(dependencies.get_db),
     allow: bool = Depends(GroupAuthorization("editor")),
 ):
+    """Add a new user to a group."""
+
     if (user_q := await db["users"].find_one({"email": username})) is not None:
         new_member = Member(user=UserOut.from_mongo(user_q))
         if (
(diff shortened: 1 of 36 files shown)


backend/app/routers/groups.py
```diff
@@ -42,30 +33,23 @@ async def save_group(
...
-    if db is None:
-        raise HTTPException(status_code=503, detail="Service not available")
-        return
     if (group := await db["groups"].find_one({"_id": ObjectId(group_id)})) is not None:
         return GroupOut.from_mongo(group)
     raise HTTPException(status_code=404, detail=f"Group {group_id} not found")
 
 
-@router.post("/{group_id}", response_model=GroupOut)
+@router.put("", response_model=GroupOut)
```
longshuicy, line 41: Any reason you leave the `group_id` out of the path parameter?

max-zilla: no that was not intentionally removed, fixed thanks

tcnichol (approved):
I created several groups. One user shared datasets only by group. The correct users can view/edit etc datasets and only the group members with roles 'editor' were able to add new group users. Everything seems to work here so marking approved.

PR #390, opened 2023-03-14
matching response type with actual dependency returns

longshuicy:
Here is a quick fix


longshuicy:
I'm wondering if the GET `/dataset/{dataset_id}/role` need to be changed similarly. Right now it's returning the whole authorization mongo object. But this endpoint is used in the test and other places... hesitant to make this change

max-zilla (approved):

tcnichol (approved):
This looks good. Approving so can be merged in.

PR #393, opened 2023-03-20
Group management UI

max-zilla:
Adds a list of groups to explore page and a simple group page listing members for each group. The latter needs a lot of work.


tcnichol (approved):
Created some groups and all the users display correctly. I'll mark this one approved.

longshuicy (approved):

PR #396, opened 2023-03-23
379 GitHub action to run codegen automatically when there is backend change

longshuicy:
I updated the readme:
```
After modifying backend API, update autogenerated client function calls:
   - Backend must be running: `docker-compose -f docker-compose.test.yml up`
   - Run codege: `npm run codegen:v2:test`
   
```
For some reason the built backend in docker container generate a very minor difference than the one we use in pycharm
image

Might be related to keycloak?


lmarini:
That endpoint `/broker/{identity_provider}/token` proxies the return of a keycloak call. All our docker file seem to fix the keycloak version to 19.0, but maybe somehow we are running a different version in one case.

longshuicy:
@lmarini i commented out the `get_idenity_provider_token` and it seems to be consistent now. I reverted the README instruction on how to run codegen. Let me know what else is left on this PR? Thanks~

backend/app/routers/keycloak.py
```diff
@@ -188,27 +188,27 @@ async def refresh_token(
...
-    else:
-        raise HTTPException(
-            status_code=status.HTTP_400_BAD_REQUEST,
-            detail={
-                "error_msg": f"Identy provider [{identity_provider}] not recognized."
-            },
-            headers={"WWW-Authenticate": "Bearer"},
-        )
+# @router.get("/broker/{identity_provider}/token")
+# def get_idenity_provider_token(
```
lmarini, line 192: @longshuicy can you add a FIXME here saying we need to parse and return a consistent response? Also can you make an issue for it? I am afraid we will forget. Thank you.

lmarini (changes requested):

lmarini (approved):

PR #391, opened 2023-03-16
Mock UI design for share dataset feature

arunapa:
* Added dataset name as parameter
* Added Share button to menu
* Added new custom component that has the share pane UI

Screenshots:
Screen Shot 2023-03-16 at 12 35 17 AM
Screen Shot 2023-03-16 at 12 35 30 AM

PR #398, opened 2023-03-27
374 add groups to datasets via gui

tcnichol:
Just adding as draft pull request to make it easier to review and get feedback.


frontend/src/components/datasets/ShareGroupDatasetModal.tsx
```diff
@@ -0,0 +1,110 @@
...
+                                labelId="demo-simple-select-label"
+                                id="demo-simple-select"
+                                label="Group"
+								defaultGroup={initial_group}
+								value={group}
+                                onChange={(event, value) => {
+                                    setGroup(event.target.value);
+                                }}
+                            >
+                            </Select>
```
tcnichol, line 82: This dropdown is always blank even though options shows all the groups added. Not sure what is wrong with it.

max-zilla:
based on #391

tcnichol:
I'm not sure what is going on with the pytest here. I ran locally and all the tests passed.

frontend/src/components/datasets/ShareGroupDatasetModal.tsx
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+
+
+	const options = Array();
+    groups.map((group) => {
+    	let group_option = {value:group.id, label:group.name}
+		options.push(group_option);
+    });
+	console.log('group optioins are', options);
```
arunapa, line 45: I've reviewed this PR and tested it, looks good to me! Two minor comments:

1. Can we remove these console log statements?
2. We need to enforce that only the dataset "owner" can share roles

I'll add the remaining changes to this branch directly so we can  merge both features together. Thanks!

frontend/src/components/datasets/ShareGroupDatasetModal.tsx
```diff
@@ -0,0 +1,142 @@
...
+									onClick={() => {
+										setShowSuccessAlert(false);
+									}}
+								>
+									<CloseIcon fontSize="inherit" />
+								</IconButton>
+							}
+							sx={{ mb: 2 }}
+						>
+                        Successfully added role!
```
arunapa, line 132: This is what the success dialog looks like
Screenshot 2023-03-31 at 7 21 00 AM

frontend/src/components/datasets/ShareGroupDatasetModal.tsx
tcnichol: We might need to think about whether an editor should be able to share a dataset. But that can be a separate issue.

max-zilla:
I tested this and the sharing feature works, two questions:

- autocomplete for users is hard-coded to "abc@xyz.com" - are we planning to update this separately? i see the group autocomplete is at least initially populated.

- when successfully shared, the pop-up box does not automatically close (see success dialog screenshot above) - is it better to close it then, or keep open to allow additional sharing?

tcnichol:
@max-zilla 

For sharing, I think it might be better to keep the box open in case a user wants to share a dataset with multiple users and groups. Though this might be something to get feedback on later. 

With the autocomplete, we might want to make any changes related to that other issues.

max-zilla (approved):

PR #400, opened 2023-03-28
Group management UI

lmarini:
Can we review and merge this as is? It has a bunch of methods to associate groups and datasets which would make it easier to test other branches, even without the ui.


lmarini:
Branch was already merged.

PR #402, opened 2023-03-29
User api keys

max-zilla:
Implements a basic functioning example of an endpoint to generate API keys and a wrapper on existing auth that lets a request use either tokens (as before) or x-api-key header.

Currently this works by:
- having a secret instance key in Clowder config.
- generating a JSON object with username + randomized token and hashing that with the instance key = the key the user receives
- when user proves the key, we use instance key to decode it back to username + token object, and lookup that in mongo to check creation time to expire if needed (defaults to 30 mins)


tcnichol (approved):
This looks good. Will mark approved and we can save improvements for later issues. We can also revisit how this will work with extractors.

PR #405, opened 2023-03-31
401 group management UI continue

longshuicy:
image
image
image

Add and delete user implemented


max-zilla (approved):
looks good and functions as expected. will make a follow-up task for a couple small improvements: 1) Add Member button is hidden if you are logged in as a Member without Editor permissions, but the delete buttons are visible. (Clicking them returns a 403 as it should). 2) Currently you can only add Members, we need a way to add Editors or change between Editor/Member.

PR #413, opened 2023-04-04
Add api key to get_current_user, catch missing auths

max-zilla:
Hotfix to address "open" endpoints - if you come in without a token or API key, the API should return a 401 rather than a server error.


tcnichol (approved):
This looks good, tested and got right error. marking approved.

PR #415, opened 2023-04-05
Sharing bugfixes

max-zilla:
Fixes a situation where a group could have multiple roles on a dataset.


tcnichol:
I merged main in so it would be easier to test.
Something I noticed (after merge, so the merge might have caused this)

1. I added Group1 to a dataset as a viewer.
2. I see a new entry in authorization where the group_id is the id for that group.
3. I use 'add group' to add Group1 as an uploader.
4. In the db, I see a new authorization entry with that group_id and the role 'uploader.'
5. The old entry was still there, but the group_ids were empty (so the permission has no effect.)

tcnichol:
Max, I did the following:
1. Added Group1 to a dataset as 'viewer.'
2. I see a new authorization object in the db with authorization 'viewer'.
3. Then I did 'share with group' and added Group1 as an 'uploader.'
4. I see a new entry for that dataset and group with 'uploader', but the old one that has 'viewer' is still there, but just with no group_ids or user_ids attached.

Not sure if this was happening before or if I accidentally messed something up when I merged main.

max-zilla:
that is intended behavior - it's ok for those lists to be empty, they aren't created until they are needed but they will be re-used if any user or group gets that role in the future.

tcnichol (approved):
Tested and will mark approved. Works as intended.

PR #416, opened 2023-04-06
WIP: 406 group management UI improvements

longshuicy:
Almost Ready to be reviewed. Need to add pytest and fix minor issues.

 In this PR those are the main things covered:

### Backend
1. modify the search endpoint to search for both name and description
2. add "editor/viewer" role option when add member
2. PUT endpoint to assign editor/viewer to member
3. Minor bugfixes related to group

### Frontend
1. Prettify the groups component to use table
2. Search interface for searching groups
3. Delete group
4. Assign editor/viewer role to group member
5. Authwrapper to properly hide/show delete, add button based on ownership 

Some of the screenshots:
image

image

image

image

### DIFF
--- a/backend/app/deps/authorization_deps.py
+++ b/backend/app/deps/authorization_deps.py
@@ -115,7 +115,7 @@ async def get_role_by_group(
         if group_out.creator == current_user:
             # Creator can do everything
             return RoleType.OWNER
-        for u in group.users:
+        for u in group_out.users:
             if u.user.email == current_user:
                 if u.editor:
                     return RoleType.EDITOR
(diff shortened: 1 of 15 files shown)


backend/app/deps/authorization_deps.py
```diff
@@ -322,7 +322,7 @@ async def __call__(
                 return True
             for u in group.users:
                 if u.user.email == current_user:
-                    if u.editor and self.role == RoleType.EDITOR:
+                    if u.editor or self.role == RoleType.EDITOR:
```
max-zilla, line 325: Unless I'm misunderstanding, I don't think this should be changed. `self.role` is the role you are requiring in the endpoint, so this would mean even if the user `u` is not an editor, if your permission requires editor this would return true for any user in the group.

longshuicy: ah i see i misunderstood the `self.role` part. Sorry about that. 
Will revert it

max-zilla (approved):

PR #421, opened 2023-04-06
fix the search

longshuicy:
Now searching when typing; when hit enter or the "search" icon, it will also search the current word. When erase or empty string, list default listeners. 

To test you can watch the network tab:
image


tcnichol (approved):
Tested and all seems to work. Marking approved.

PR #425, opened 2023-04-07
job created timestamp wrong - fixed to local time

tcnichol:
Looks like 'created' used utcnow instead of now. This is changed and now extractor logs seem consistent on time.
Screen Shot 2023-04-07 at 2 11 34 PM


longshuicy (approved):
Looks correct now

PR #426, opened 2023-04-10
implement expires parameter for user keys

max-zilla:
Allow dynamic expiration of individual API keys.

To generate a key, POST to /users/keys?mins=2 for a 2-minute expiration, default is 30.


tcnichol (approved):
This one seems to work fine. I got the new key and am able to set different times for expiration.

PR #428, opened 2023-04-11
Show Creator on Group Page

tcnichol:
The owner is now shown in the group table. Instead of 'editor' or 'member' the owner says 'owner' and that role cannot be changed. I noticed that you could create a group and then change your own role to 'member' but that you could still edit roles, including your own. 

I would like to figure out a way to get the owner listed first, which is why this is a draft pull request.


tcnichol:
Added a line on top under the description that lists the group creator.
I notice that for groups, there is only 'creator' and 'editor' and 'member' and that there is no actual role called 'owner' and that it seems like 'creator' is always uniquely the same person. 

A few things to test in the future - editors can change editors to members. And it looks like an editor can change their own role to member. This creates a possible situation where all editors make themselves members instead. The creator now can't be changed, but we might want to think about blocking users from downgrading their own roles.

frontend/src/components/groups/MembersTableUserEntry.tsx
```diff
@@ -61,74 +62,102 @@ export function MembersTableUserEntry(props: MembersTableUserEntryProps) {
...
-								<MenuItem value={"member"}>Member</MenuItem>
-							  </Select>
-							</FormControl>
-						:
-						member.editor !== undefined && member.editor ?
-							"Editor" : "Member"
-				}
-				{/*only owner or editor are allowed to modify roles of the member*/}
-				<AuthWrapper currRole={role} allowedRoles={["owner", "editor"]}>
+	console.log(member, 'member')
```
lmarini, line 65: Can we remove?

frontend/src/components/groups/MembersTableUserEntry.tsx
```diff
@@ -61,74 +62,102 @@ export function MembersTableUserEntry(props: MembersTableUserEntryProps) {
...
+
+	if (member.user.email == creatorEmail) {
```
lmarini, line 67: Since all we are doing is not letting the user change this, can we put the if statement around only those controls and not the entire row? This way we don't have to duplicate as much code?

lmarini (changes requested):

frontend/src/components/groups/Group.tsx
```diff
@@ -92,6 +93,7 @@ export function Group() {
 							<RoleChip role={role}/>
 						</Box>
 						<Typography variant="body1" paragraph>{groupAbout.description}</Typography>
+						<Typography variant="body1" paragraph>{groupAbout.creator} is the group creator</Typography>
```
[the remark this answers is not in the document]
tcnichol: Sure. Will change.

frontend/src/components/groups/MembersTableUserEntry.tsx
tcnichol: will remove.

frontend/src/components/groups/Group.tsx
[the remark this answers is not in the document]
tcnichol: I sort of have this changed, but now the creator: and the actual email of the creator seem to be broken into 2 lines. I'm trying to figure out how to fix this. But if anyone sees an easy fix, let me know.

frontend/src/components/groups/MembersTableUserEntry.tsx
tcnichol: Put each TableCell into conditionals. This also makes it clearer what is different based on Owner.

frontend/src/components/groups/Group.tsx
```diff
@@ -92,6 +96,9 @@ export function Group() {
 							<RoleChip role={role}/>
 						</Box>
 						<Typography variant="body1" paragraph>{groupAbout.description}</Typography>
+						<Typography variant="body1" paragraph><strong>Creator: </strong><ListItem>
+   							<a href={groupCreatorEmailLink} target="_top">{groupCreatorEmail}</a>
```
longshuicy, line 100: image
Instead of plain <a> Tag, can we use either MUI <Link> tag so it turns orange :-)

longshuicy:
Just a thought, if it's easy to do maybe we can filter out the owner's email as well when adding members
image

longshuicy (approved):

lmarini (approved):

PR #430, opened 2023-04-13
Sharing tab cleanup

max-zilla:


backend/app/routers/authorization.py
```diff
@@ -232,7 +230,23 @@ async def set_dataset_user_role(
...
+                    async for group_q in group_q_list:
+                        group = GroupOut.from_mongo(group_q)
+                        for u in group.users:
+                            if u.user.email == username:
+                                group_occurrences += 1
+                    if auth_db.user_ids.count(username) == group_occurrences:
+                        auth_db.user_ids.append(username)
+                        await db["authorization"].replace_one(
+                            {"_id": auth_db.id}, auth_db.to_mongo()
+                        )
```
longshuicy, line 248: I'm out of loop and wondering what's this "counting user in all groups" logic is for? Could you kindly explain just a little bit? Thanks!

backend/app/routers/authorization.py
```diff
@@ -336,65 +350,50 @@ async def remove_dataset_user_role(
         raise HTTPException(status_code=404, detail=f"Dataset {dataset_id} not found")
 
 
-@router.get("/datasets/{dataset_id}/users_and_roles", response_model=List[UserAndRole])
-async def get_dataset_users_and_roles(
+@router.get("/datasets/{dataset_id}/roles")  # , response_model=DatasetRoles
```
longshuicy, line 353: if you don't need the `response_model` maybe just remove it?

frontend/src/components/sharing/GroupAndRoleTableEntry.tsx
```diff
@@ -0,0 +1,157 @@
...
+					{group_role.group.users ? group_role.group.users.length : 0}
+					<IconButton
+						type="button"
+						sx={{ p: "10px" }}
+						aria-label="edit"
+						onClick={() => {
+							setExpand(!expand);
+						}}
+					>
+						<ListIcon sx={iconStyle} />
```
longshuicy, line 87: nitpicking... maybe an arrow up and arrow down icon would be more intuitive to indicate that user can click?

backend/app/routers/authorization.py
max-zilla: Each time the dataset is shared with a user or a group containing the user, that user's email is added to the list of auth_db.user_ids. This means an email might appear multiple times: <number of groups user is in> + <one occurrence for the user themselves>.

If you share the dataset with the same user repeatedly, say as a viewer, we dont want to add multiple occurrences of their email to the list, only once. BUT the user's email might be in the list 4 times thanks to being in 4 groups, but doesn't yet have an entry for their individual self. 

So we need to figure out if all the entries in the list come from groups, in which case we can add a new entry, OR if there are only n-1 entries from groups and the last entry already exists, in which case we don't add a new entry.

backend/app/routers/authorization.py
max-zilla: added back thanks, it is accurate so not sure why i had it commented out

backend/app/routers/authorization.py
longshuicy: Thanks for the explanation that makes sense. I was wondering why you evaluate `auth_db.user_ids.count(username) == group_occurrences` now it makes sense.

longshuicy (approved):

longshuicy:
Bug resolved. Going to merge when github actions finish

PR #445, opened 2023-04-19
441 gui for create new group

longshuicy:
Simple UI to create group leverage rjsf library. Lint change is following the "action on save" config on pycharm

image

image


max-zilla (approved):
everything works well

PR #451, opened 2023-04-20
Fix reload bug

max-zilla:
When refreshing dataset page, was getting a JWTError that wasn't being caught in refresh_token(). resulted in being bounced back to main page every time. added this to existing exceptions, seems to resolve behabior.


longshuicy (approved):
Looks good

PR #459, opened 2023-04-20
Autocomplete on sharing pop-ups

max-zilla:
This implements Autocomplete form with dynamic state for both user and group sharing forms on dataset page. Groups were already there, but switched to Autocomplete as code was cleaner and we should use that pattern where possible.


longshuicy (approved):
Looks good approving

PR #468, opened 2023-04-25
461 fix library version in pipfile

longshuicy:
To avoid future github action failure, i'm fixing all the library versions. 
I'm using the version that works on my machine, not necessarily the latest version.
Let me know if you'd prefer a later version


lmarini (approved):
Didn't run it, but thank you for doing this! This is great.

PR #470, opened 2023-04-26
Upgrade minio rabbitmq related in helm chart

longshuicy:
For now solution is
1. upgrade mino to v12 which supports versioning in standalone mode
2. decrease the replica to 1
3. stay with nfs-taig

other changes on this PR:
- add clusterIp
- one replica of minio doesn't need headless services (?) so change the host to just clowder2-minio


tcnichol:
This looks good, aside from running it locally what testing should be done?

longshuicy:
> This looks good, aside from running it locally what testing should be done?

It's deployed here https://clowder2.software-dev.ncsa.cloud/

PR #471, opened 2023-04-26
469 create profile page

tcnichol:
Added a basic profile page.
Right now since admin isn't enabled, the admin field is always false.
Does need some improvement as far as looks, so this is a draft for now.


longshuicy (approved):
Looks good. Do you want to create new issues to document how we can further populate the profile page?

PR #478, opened 2023-05-01
Fixed bug where error detail might be empty

arunapa:
* Added null check to the detail string to prevent errors when detail is not set

### DIFF
--- a/frontend/src/actions/common.js
+++ b/frontend/src/actions/common.js
@@ -64,7 +64,7 @@ export function handleErrors(reason, originalFunc){
 		return (dispatch) => {
 			dispatch({
 				type: NOT_AUTHORIZED,
-				reason: reason.body.detail,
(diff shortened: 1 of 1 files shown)


frontend/src/actions/common.js
```diff
@@ -64,7 +64,7 @@ export function handleErrors(reason, originalFunc){
 		return (dispatch) => {
 			dispatch({
 				type: NOT_AUTHORIZED,
-				reason: reason.body.detail,
+				reason: reason.body.detail !== undefined? reason.body.detail : "Forbidden",
```
longshuicy, line 67: could you even try check if `body` is a field of reason by doing `reason.body !== undefined && reason.body.detail !== undefined ? ...`?

arunapa: Makes sense, I'll modify the code. Thanks ~

longshuicy (approved):

PR #482, opened 2023-05-03
format in local timezone

longshuicy:
Because backend datetime.utcnow() doesn't come with a trailing Z (Zulu), the frontend cannot treat the time as utc. 
The fix would be on the frontend add Z and parse with timezone info.

This explains why python datetime.utcnow() doesn't have the Z https://stackoverflow.com/a/63731605

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -25,6 +25,7 @@
 				"babel-polyfill": "^6.26.0",
 				"classnames": "^2.2.6",
 				"date-fns": "^2.28.0",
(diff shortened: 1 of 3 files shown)


max-zilla:
Should we consider using `datetime.now(timezone.utc)` on backend to get the Z in all our objects without the frontend adding?

tcnichol (approved):
Checked this one, marking approved.

tcnichol:
@max-zilla that might be a good idea, since i'm thinking with this fix the back end and front end might be showing inconsistent datetimes.

PR #487, opened 2023-05-04
438 browser back on 403 page takes us to the dataset page we dont have access to

arunapa:
https://user-images.githubusercontent.com/28983148/236076643-a35a3553-991c-4854-adf3-10865b7c11b5.mov


tcnichol:
Here is what I see currently happening. I get an error pop up, and back takes me to explore. Is

https://github.com/clowder-framework/clowder2/assets/40038535/db718e01-5636-4e1f-b84e-e8a13da84600

 this what should be happening?

tcnichol (approved):
After meeting we agree that what we see is intended behavior, marking approved.

PR #488, opened 2023-05-04
486 list doesnt get populated immediately when successfully submitting share permissions

arunapa:
https://user-images.githubusercontent.com/28983148/236089767-dd4d3aac-3217-499b-8ed2-0ebb863e36fd.mov

### DIFF
--- a/frontend/src/components/datasets/ShareDatasetModal.tsx
+++ b/frontend/src/components/datasets/ShareDatasetModal.tsx
@@ -19,7 +19,7 @@ import {
 	Typography,
 } from "@mui/material";
 import { useParams } from "react-router-dom";
-import { setDatasetUserRole } from "../../actions/dataset";
+import { fetchDatasetRoles, setDatasetUserRole } from "../../actions/dataset";
 import { useDispatch, useSelector } from "react-redux";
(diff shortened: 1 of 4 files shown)


frontend/src/components/datasets/ShareDatasetModal.tsx
```diff
@@ -46,9 +46,12 @@ export default function ShareDatasetModal(props: ShareDatasetModalProps) {
 	const [options, setOptions] = useState([]);
 	const users = useSelector((state: RootState) => state.group.users);
 
-	const setUserRole = (datasetId: string, username: string, role: string) =>
+	const setUserRole = async (datasetId: string, username: string, role: string) =>
```
longshuicy, line 49: oh why is the `async` needed

arunapa: To make sure that when the subsequent listRoles is called, the update is already reflected in the backend

longshuicy: ah sorry i didn't realize you are directly using dispatch :-) My bad it makes perfect sense now.

longshuicy (approved):
works well approved

max-zilla (approved):

PR #504, opened 2023-05-16
Beanie improvements (id, mongoclient, file calls)

max-zilla:
- remove ID fields from beanie documents and try to use auto-casting to _Out classes using class.dict()
- remove get_db dependency and remaining references to db[] calls using MongoClient
- replace remaining calls to db[files] db[file_versions] db[folders] and others
- implement file and folder views in Beanie format

Many aspects are untested, but wanted to get these fixes in.

### DIFF
--- a/backend/app/rabbitmq/heartbeat_listener_sync.py
+++ b/backend/app/rabbitmq/heartbeat_listener_sync.py
@@ -58,7 +58,7 @@ def callback(ch, method, properties, body):
             processed_feed = await _process_incoming_v1_extractor_info(
                 extractor_name, extractor_out.id, process
             )
-            processed_feed.insert()
+            await processed_feed.insert()
 
         return extractor_out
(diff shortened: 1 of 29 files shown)


backend/app/rabbitmq/heartbeat_listener_sync.py
```diff
@@ -62,7 +58,7 @@ def callback(ch, method, properties, body):
             processed_feed = await _process_incoming_v1_extractor_info(
                 extractor_name, extractor_out.id, process
             )
-            db["feeds"].insert_one(processed_feed)
+            processed_feed.insert()
```
longshuicy, line 61: wonder if we need await here?

max-zilla: good catch, added this and a few other missing awaits also

PR #514, opened 2023-05-31
Use dedicated test DB

max-zilla:
Uses a test database for pytest instead of the main DB.

Adds extractor update message listener to prod dockerfile.

Also will look for X-API-KEY (all caps) in authorization by key now.


longshuicy (approved):
Ran a few pytest and see it populates the test database. Approving.

longshuicy (approved):
Ran a few pytest and see it populates the test database. Approving.

PR #515, opened 2023-05-31
507 create datasets form improvements

longshuicy:
Made changes according to the description section in the issue.
I realize we can overwrite the default error prompt so I replaced it with the MUI alert component; also removed the "." in the error message.

image


max-zilla (approved):

max-zilla:
please make description a text box / taller box.

lmarini (approved):

PR #522, opened 2023-06-02
521 explore code splitting and lazy loading

longshuicy:
Update:
1. give code split name by passing in this webpackChunkName when importing
```
const Demo = lazy(
	() => import(/* webpackChunkName: "previewers-demo" */ "../previewers/Demo")
);
```

2. reconfigure webpack to give common vendor chunk name

image

There is still this bundle with id that i can't figure out why
---------
This is an experiment PR that has a demonstrative component that uses vega to generate some visualization. You can trigger this by going to **Dataset** --> **Preview Tab**
test

### Lazy loading
Only support exportation of **default** component for some reason. I have to write:
```
export default function Demo()
```
instead of 
```
export const Demo = () =>
```

Import looks like this
```
const Demo = lazy(() => import("../previewers/Demo"));

```

### Multiple bundles
index html

### Using Suspense to transitioning
Pasted Graphic 9

### Write custom error boundary
ONLY support HOC component; not the functional component
https://legacy.reactjs.org/docs/error-boundaries.html#introducing-error-boundaries
Error boundaries work like a JavaScript catch {} block, but for components. Only class components can be error boundaries. In practice, most of the time you’ll want to declare an error boundary component once and use it throughout your application.
Pasted Graphic 5


longshuicy:
closing this PR. Everything is covered in #557

PR #527, opened 2023-06-05
Error running extractor on main

tcnichol:
Fix was that file.id was cast to string but not the datasetid.


lmarini (approved):
Worked great.

PR #528, opened 2023-06-05
fix extractor error beanie branch

tcnichol:
Like main, the dataset_id and folder_id need to be str and not objectid


ddey2 (changes requested):
Doesn't work still. shows backend failure.

TypeError: 'DatasetDB' object is not subscriptable

ddey2:
Screenshot 2023-06-05 at 3 13 44 PM

tcnichol:
I will clear my db and check again on this.

lmarini:
@tcnichol said this might have started on the wrong branch. He will take a second look.

max-zilla:
I believe this particular bug (and some associated ones) are fixed in my PR here: https://github.com/clowder-framework/clowder2/pull/534
we can merge this first or just close it if you prefer.

PR #529, opened 2023-06-05
fix filter by status

longshuicy:
image

put those status in the config file instead of hardcoding in the component


ddey2 (approved):
Looks good.

lmarini (approved):

PR #532, opened 2023-06-05
525 should show error message on frontend when user tries to add an unknown user to groups

longshuicy:
This PR:
1. wrap common code to util class
2. add missing error message dialog to `/groups` page and `/group/{id}` page

To test:
1. Go to any existing group page and click add member
2. Type random strings to the email 
image

3. see if there is error modal
image

### DIFF
--- a/frontend/src/components/Explore.tsx
+++ b/frontend/src/components/Explore.tsx
@@ -1,19 +1,19 @@
-import React, {useEffect, useState} from "react";
-import {Box, Button, ButtonGroup, Grid, Tab, Tabs} from "@mui/material";
+import React, { useEffect, useState } from "react";
+import { Box, Button, ButtonGroup, Grid, Tab, Tabs } from "@mui/material";
 
-import {Dataset, RootState} from "../types/data";
-import {useDispatch, useSelector} from "react-redux";
-import {fetchDatasets} from "../actions/dataset";
-import {resetFailedReason} from "../actions/common";
-import {downloadThumbnail} from "../utils/thumbnail";
+import { Dataset, RootState } from "../types/data";
+import { useDispatch, useSelector } from "react-redux";
+import { fetchDatasets } from "../actions/dataset";
+import { resetFailedReason } from "../actions/common";
+import { downloadThumbnail } from "../utils/thumbnail";
 
-import {a11yProps, TabPanel} from "./tabs/TabComponent";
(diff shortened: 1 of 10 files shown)


lmarini (changes requested):
Works well. Two potential requests:

1. Could we move the error dialogs to their own componets? While the component iteself doesn't take up much space, with the events and everything else it adds a lot of complexity to the parent component that could be encapsulated in its own component
2. Why not just print an error above/below the text field? The error dialog is more for unexpected errors. In this case it is more of a validation issue.

longshuicy:
> Works well. Two potential requests:
> 
> 1. Could we move the error dialogs to their own componets? While the component iteself doesn't take up much space, with the events and everything else it adds a lot of complexity to the parent component that could be encapsulated in its own component
> 2. Why not just print an error above/below the text field? The error dialog is more for unexpected errors. In this case it is more of a validation issue.

if we are not doing a popup then definitely we can move the logic to its own component. Let me try a bit

tcnichol:
I checked and this works.

longshuicy:
> Works well. Two potential requests:
> 
> 1. Could we move the error dialogs to their own componets? While the component iteself doesn't take up much space, with the events and everything else it adds a lot of complexity to the parent component that could be encapsulated in its own component
> 2. Why not just print an error above/below the text field? The error dialog is more for unexpected errors. In this case it is more of a validation issue.

1 is done. Now there is a ErrorModal component. 
However 2 is tricky... I did create an inline alert component, but there is no good way that we specify "this error will trigger dialog error modal" vs "this error should be an inline alert"; because those two are triggered by the same "error" state that essentially are error status returned by the backend

tcnichol (approved):
Re-ran and all tests passed, looks like all comments were addressed and works so marking this approved.

max-zilla (approved):

longshuicy:
Todos:
1. add an search endpoint matches the keywords search on type so user get to be matched quickly
2. form validation for email
3. write specific error handler and implement inline alert listening for that action
4. write some documentations about how we handle errors (new task)

ddey2 (approved):
Looks so good, better than previous one. Works well. 

One suggestion, does the group icon need to be before the group name?
Screenshot 2023-06-09 at 5 00 35 PM

lmarini (approved):

PR #535, opened 2023-06-06
download file version button hard to see

longshuicy:
image

It doesn't look the best but the button is more obvious now


tcnichol (approved):
This looks good, button is bigger and more visible and marking approved.
Saw some tests failed so re-running tests.

max-zilla (approved):

PR #557, opened 2023-06-20
549 example of viz component

longshuicy:
In this PR:

1. Load widgets at the root `visualization.config.tsx`
2. a few example preview components as well as their manifest.json are set


frontend/src/components/previewers/Audio/manifest.json
```diff
@@ -0,0 +1,20 @@
...
+	"main": "Audio.tsx",
+	"dependencies": {
+		"clowder2-core": "1.0.0",
+		"react": "^17.0.2",
+		"react-dom": "^17.0.2"
+	},
+	"vizConfig": {
+		"name": "Audio",
+		"mainType": "audio",
+		"mimeType": [
```
sandeep-ps, line 14: Since this is a list, would using the plural form (`mimeTypes`) maybe better? What do you think?

tcnichol (approved):
Looks good. I'm marking approved.

sandeep-ps (approved):
I tested this PR with image, video, audio, and text files, and the previews look good to me. Approving. Thanks.

PR #559, opened 2023-06-21
556 viz config for files

tcnichol:
Here is a sample to send in as a json for creating a viz config object:

`{
"resource":{
    "resource_id":"649496ed198de6c57f3e8860",
    "collection":"files"
},
"client":"testClient",
"viz_config_data":{"url":"someurl.com"},
"visualization":"649496ed198de6c57f3e8860",
"visualization_mimetype":"testMimeType"}`

There are also fields for extractor_info and job if the viz config is generated from an extractor, but here I used the 'client' field with just a placeholder value. 

The viz config data is for vizualizations like the WMS openlayers, where the value would be a URL.


backend/app/routers/files.py
```diff
@@ -409,3 +410,43 @@ async def resubmit_file_extractions(
             FileOut(**file.dict()), rabbitmq_client, user, credentials
         )
     return resubmit_success_fail
+
+@router.post("/{file_id}/visualization")
+async def save_visualization(
```
ddey2, line 415: can you rename this to visualization_config, may be? I named visualization in my PR for the data

backend/app/routers/files.py
```diff
@@ -409,3 +410,43 @@ async def resubmit_file_extractions(
             FileOut(**file.dict()), rabbitmq_client, user, credentials
         )
     return resubmit_success_fail
+
+@router.post("/{file_id}/visualization")
```
sandeep-ps, line 414: Since this is a collection resource (i.e., many visualizations for a resource), it may be better to use the plural form (visualizations) in the URL. What do you think?

backend/app/routers/files.py
tcnichol: Everything should be renamed now. 

These will all be in separate routes in visualizations router.

backend/app/routers/files.py
tcnichol: renamed.

backend/app/routers/files.py
ddey2: I see the separate router. looks good. Do you still need these in files.py?

tcnichol: No, I will remove them.

backend/app/routers/visualization.py
```diff
@@ -0,0 +1,129 @@
...
+    user=Depends(get_current_user),
+    credentials: HTTPAuthorizationCredentials = Security(security),
+):
+    if (viz_config := await VizConfigDB.get(PydanticObjectId(config_id))) is not None:
+        return viz_config.dict()
+    else:
+        raise HTTPException(status_code=404, detail=f"VizConfig {config_id} not found")
+
+
+@router.patch("/config/{config_id}/map", response_model=VizConfigOut)
```
longshuicy, line 104: just curious what's this endpoint for?

backend/app/models/viz_config.py
```diff
@@ -0,0 +1,38 @@
...
+from app.models.authorization import AuthorizationDB
+from app.models.listeners import ExtractorInfo, EventListenerJobDB
+from app.models.metadata import MongoDBRef
+from app.models.pyobjectid import PyObjectId
+from app.models.users import UserOut
+
+
+class VizConfigBase(BaseModel):
+    resource_id: PydanticObjectId
+    resource_type: str
```
longshuicy, line 16: Since we are using the resource ref on the VizConfigDB, do we still need the `resource_id` and `resource_type`?

backend/app/routers/visualization.py
tcnichol: this is for when the vizualization is something like a URL - the way that an openlayers GIS preview works. It probably needs a better name or maybe I can just call it vizualization_data?

The map was because it should be key value pairs.

backend/app/models/viz_config.py
tcnichol: I'll remove these, you are right they are no longer needed.

backend/app/main.py
```diff
@@ -167,6 +169,12 @@
     tags=["groups"],
     dependencies=[Depends(get_current_username)],
 )
+api_router.include_router(
+    visualization.router,
+    prefix="/visualization",
```
lmarini, line 174: We have been using plurals for the custom routes. That's usually the case for rest.

backend/app/models/viz_config.py
```diff
@@ -0,0 +1,34 @@
...
+from pydantic import Field, BaseModel
+
+from app.models.authorization import AuthorizationDB
+from app.models.listeners import ExtractorInfo, EventListenerJobDB
+from app.models.metadata import MongoDBRef
+from app.models.pyobjectid import PyObjectId
+from app.models.users import UserOut
+
+
+class VizConfigBase(BaseModel):
```
lmarini, line 14: Can we spell out Visualization?

backend/app/routers/visualization.py
```diff
@@ -0,0 +1,127 @@
...
+            viz_config = VizConfigDB(**vizconfig_in, resource=resource_ref)
+            await viz_config.insert()
+            return viz_config.dict()
+        else:
+            raise HTTPException(
+                status_code=404, detail=f"Dataset {resource_id} not found"
+            )
+
+
+@router.get("/{file_id}/config_file", response_model=List[VizConfigOut])
```
lmarini, line 58: Can we drop the `config_file` part and maybe combine the three routes below? I don't think we need to know what resource it is attached to in the route. For example `/visualizations/123/config_file`. If you are just checking if the resource exists, since the model has the mongo reference, you could use that to know which collection to query.

tcnichol: I now have just one route that checks for these.

backend/app/routers/visualization.py
```diff
@@ -0,0 +1,118 @@
...
+            viz_config = VisualizationConfigDB(**vizconfig_in, resource=resource_ref)
+            await viz_config.insert()
+            return viz_config.dict()
+        else:
+            raise HTTPException(
+                status_code=404, detail=f"Dataset {resource_id} not found"
+            )
+
+
+@router.get("/{file_id}/config", response_model=List[VisualizationConfigOut])
```
longshuicy, line 62: This is only for file?

longshuicy: what about dataset?
could we make it {resource_id}/config?

longshuicy: in the logic you probably need to check if it's dataset or file

longshuicy:
All endpoints works. Will approve after the only comment i have about the endpoint `@router.get("/{file_id}/config", response_model=List[VisualizationConfigOut])`

longshuicy (approved):

ddey2 (approved):

PR #560, opened 2023-06-22
Implement backend for visualization data

ddey2:


backend/app/models/visualizations.py
```diff
@@ -0,0 +1,46 @@
...
+
+class VisualizationsDB(View, VisualizationsBase):
+    creator: UserOut
+    created: datetime = Field(default_factory=datetime.utcnow)
+    modified: datetime = Field(default_factory=datetime.utcnow)
+    bytes: int = 0
+    content_type: FileContentType = FileContentType()
+
+    class Settings:
+        name = "Visualizations"
```
longshuicy, line 33: maybe lowercase for the collection name just follow other conventions?

backend/app/models/visualizations.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+class VisualizationsList(View, VisualizationsBase):
```
lmarini, line 36: I don't think we need this, unless we need a view to link visualization configs to permissions. In the case of datasets and files, we are using mongo views to include authorization information. Here we are not doing that. For example:

```
    class Settings:
        source = DatasetDB
        name = "datasets_view"
        pipeline = [
            {
                "$lookup": {
                    "from": "authorization",
                    "localField": "_id",
                    "foreignField": "dataset_id",
                    "as": "auth",
                }
            },
        ]
```

backend/app/models/visualizations.py
```diff
@@ -0,0 +1,46 @@
...
+from typing import Optional, List
+
+from beanie import Document, View, PydanticObjectId
+from pydantic import BaseModel, Field
+
+from app.models.files import FileContentType
+from app.models.users import UserOut
+
+
+class VisualizationsBase(BaseModel):
```
lmarini, line 11: We use singular for other models. Also maybe we should call this `VisualizationConfig` to be more specific.

backend/app/models/visualizations.py
ddey2: done

backend/app/models/visualizations.py
ddey2: agree. I removed this.

backend/app/routers/visualization.py
```diff
@@ -0,0 +1,85 @@
...
+from app.config import settings
+from app.keycloak_auth import get_current_user
+
+from app.models.visualization import VisualizationOut, VisualizationIn, VisualizationDB
+from app.routers.utils import get_content_type
+
+router = APIRouter()
+
+
+@router.post("/name={name}&description={description}", response_model=VisualizationOut)
```
longshuicy, line 19: in a post endpoint maybe it's better to pass in the variables as body instead of query parameters? Thoughts?

longshuicy (commented):
Pytest passes and endpoints looks good.
I'm wondering from the client point of view we might need an endpoint to get the bytes of the viz data? (or is it somehow already covered?)

tcnichol (approved):
Looks good. I will mark approved since all the codegen stuff is now fixed.

backend/app/routers/visualization.py
```diff
@@ -0,0 +1,106 @@
...
+    ) is not None:
+        fs.remove_object(settings.MINIO_BUCKET_NAME, visualization_id)
+        visualization.delete()
+        return
+    raise HTTPException(
+        status_code=404, detail=f"Visualization {visualization_id} not found"
+    )
+
+
+@router.get("/download/visualization_id={visualization_id}")
```
longshuicy, line 87: could this route just be `/download/{visualization_id}`

backend/app/routers/visualization.py
longshuicy: Could you try remove name={name}&description={description} in the path and see if fastAPI can still pick it up as query paratmers

longshuicy (approved):

PR #561, opened 2023-06-23
Remove advanced search toggle

max-zilla:
Because lucene search is not properly passing authorization to backend, we need to disable.


longshuicy:
Looks good approve

longshuicy (approved):

PR #564, opened 2023-06-23
Fix pytest failing upgrade cache action

longshuicy:


tcnichol (approved):
This fixes the problem, marking approved.

ddey2 (approved):
works.

Issue #457, opened 2023-04-20
Switching PROD/DEV docker compose requires to deleted keycloak volume


lmarini:
When we switching from prod docker compose to dev docker compose, it's still using the original keycloak volume.
Check if anything is being cached?

to reproduce login, and check the path to the redirection

lmarini:
Looks like whatever docker compose gets run first, that `KC_DB_URL_DATABASE` gets used to create the container. When the other docker compose is used to start the services, if the postgres volume is already there, it ignores it and doesn't create the other postgres database. So if you start with the production docker compose, you get a `keycloak_prod` created. But when you switch to the dev docker compose, no `keycloak_dev` is created. I tested this by going into the terminal and trying to connect to both databases with `psql -h localhost -U keycloak -d keycloak_dev` and `psql -h localhost -U keycloak -d keycloak_prod`.

lmarini:
Found reasoning. Makes sense. 

> Warning: the Docker specific variables will only have an effect if you start the container with a data directory that is empty; any pre-existing database will be left untouched on container startup.

https://hub.docker.com/_/postgres

longshuicy:
addressed in #565

PR #565, opened 2023-06-23
Set the docker project name to ``-p clowder2-dev` when running the dev

lmarini:
stack so that keycloak doesn't fail if both dev and prod are ran on the same host. #457


ddey2 (approved):

longshuicy (approved):
Merging when github actions succeed

PR #567, opened 2023-06-26
only files not in folders show up at root level

tcnichol:
Only files that are not in folders show up in root level, files also show up in proper folders.


longshuicy (approved):
Works. Approved

PR #571, opened 2023-06-27
570 api key failed on frontend

longshuicy:
Fix the API key page and deletion


tcnichol (approved):
Tested. Can create and delete api keys no error in console.

Pipfile
```diff
@@ -0,0 +1,11 @@
+[[source]]
```
lmarini, line 1: Is this file in the root accidental?

lmarini (changes requested):

PR #573, opened 2023-06-30
572 frontend visualization reading from viz config and viz data

longshuicy:
This PR uses Visualization config and Visualization data infrastructure of the backend to render visualization. 
- In the case that there is viz config associate with the resource, use the field `component_name` and `visualization` to instantiate matching component and viz bytes; 
- If there is NO viz config, fallback to streaming the raw file bytes by guess which widget to use.

Technical Dive into

To test. Note the below steps will eventually being taken care of by extractors. Not meant for user to do those.

1. Create a dataset and upload a **powerpoint** or other complicated file type that current no widget supports. Write down the resource id.
2. Create a viz data (e.g.) associate with that resource. e.g.
POST http://localhost:8000/api/v2/visualizations?name=thumbnail&description=thumbnail with `form/data` body
image

You should get back the json
```
{
    "name": "thumbnail",
    "description": "thumbnail",
    "id": "649eff6454dfbba7f6375841",
    "creator": {
        "email": "cwang138@illinois.edu",
        "first_name": "Chen",
        "last_name": "Wang",
        "id": "649d989c14c164e28d8c7113"
    },
    "created": "2023-06-30T16:14:28.900698",
    "modified": "2023-06-30T16:14:28.900702",
    "bytes": 626589,
    "content_type": {
        "content_type": "image/png",
        "main_type": "image"
    }
}
```
3. Register the viz config. Make sure you put correct `resource_id`, `visualization` and  `component_name`. Visualization is the id of the viz data. e.g.
POST http://localhost:8000/api/v2/visualizations/config
```
{
    "resource": {
        "resource_id": "649effa354dfbba7f6375842",
        "collection": "files"
    },
    "client": "testClient",
    "viz_config_data": {},
    "visualization": "649eff6454dfbba7f6375841",
    "component_name": "basic-image-component"
}
```
4. now you should be able to visualize that file that originally has no visualization


backend/app/models/visualization_config.py
```diff
@@ -17,6 +13,8 @@ class VisualizationConfigBase(BaseModel):
     job: Optional[EventListenerJobDB]
     client: Optional[str]
     viz_config_data: dict = {}
+    visualization: str
```
lmarini, line 16: Can we rename this to `visulization_id` to make it explicit that it is a reference to an ObjectId? I think that's the common practice. Also should the type be PyObjectId?

Eventaually this should probably be `visualization_data_id` or something similar to be more specific (see slack comment).

backend/app/models/visualization_config.py
```diff
@@ -17,6 +13,8 @@ class VisualizationConfigBase(BaseModel):
...
+    component_name: str
```
lmarini, line 17: Maybe change to `visualization_component_id` or something similar to be more specific?

frontend/src/actions/visualization.js
```diff
@@ -0,0 +1,120 @@
+import { V2 } from "../openapi";
+import { handleErrors } from "./common";
+import config from "../app.config";
+import { getHeader } from "../utils/common";
+
+export const GET_VIZ_CONFIG = "GET_VIZ_CONFIG";
```
lmarini, line 6: I see a lot of `viz` instead of `vis` in the code base. I am leaning towards `vis` but I would love to hear what others think.

frontend/src/actions/visualization.js
```diff
@@ -0,0 +1,120 @@
...
+	visualizationId,
+	filename = "",
+	autoSave = true
+) {
+	return async (dispatch) => {
+		if (filename === "") {
+			// TODO guess extension
+			filename = `${visualizationId}.tmp`;
+		}
+		let endpoint = `${config.hostname}/api/v2/visualizations/download/${visualizationId}`;
```
lmarini, line 58: I know this endpoint was added in a different PR, but I was wondering if the following would be a better URL:

`/visualizations/${visualizationId}/bytes` or `/visualizations/${visualizationId}/blob`

@ddey2

frontend/src/actions/visualization.js
```diff
@@ -0,0 +1,120 @@
...
+		const response = await fetch(endpoint, {
+			method: "GET",
+			mode: "cors",
+			headers: await getHeader(),
+		});
+
+		if (response.status === 200) {
+			const blob = await response.blob();
+			if (autoSave) {
```
lmarini, line 67: How is this used?

frontend/src/components/visualizations/Audio/Audio.tsx
```diff
@@ -0,0 +1,74 @@
...
+	fileId?: string;
+	visualizationId?: string;
+};
+
+export default function Audio(props: AudioProps) {
+	const { fileId, visualizationId } = props;
+
+	const dispatch = useDispatch();
+
+	const generateFileDownloadUrl = (
```
lmarini, line 24: I wonder if there is a way to simplify this set of chained effects and actions. We can chat about this in person.

frontend/src/components/visualizations/Visualization.tsx
```diff
@@ -0,0 +1,83 @@
...
+											return React.cloneElement(
+												vizComponentDefinition.component,
+												{
+													visualizationId: vizConfigEntry.visualization,
+												}
+											);
+										}
+									});
+								}
+								// if no vizaulization config exist, guess which widget to use by looking at the mime type of
```
lmarini, line 52: typo: vizaulization

frontend/src/components/visualizations/Visualization.tsx
```diff
@@ -0,0 +1,83 @@
...
+		dispatch(getVizConfigAction(resourceId));
+
+	useEffect(() => {
+		listFileSummary(fileId);
+		getVizConfig(fileId);
+	}, []);
+
+	return (
+		<>
+			{vizComponentDefinitions.map((vizComponentDefinition) => {
```
lmarini, line 34: This block might use a few more comments. Lots of steps going on that might not be clear.

frontend/src/components/visualizations/Visualization.tsx
```diff
@@ -0,0 +1,83 @@
...
+				return (
+					<LazyLoadErrorBoundary fallback={<div>Fail to load...</div>}>
+						<Suspense fallback={<div>Loading...</div>}>
+							{(() => {
```
lmarini, line 38: Is this anonymous function needed here? Does Suspense require a function?

lmarini (changes requested):
A few minor comments. Haven't tested it yet.

tcnichol:
I did test and works.

frontend/src/actions/visualization.js
tcnichol: No strong opinion either way, maybe leaning slightly towards VIS?

frontend/src/actions/visualization.js
ddey2: @lmarini I will update the url in a separate task.

longshuicy: will be done in #580

frontend/src/actions/visualization.js
longshuicy: This `autoSave` is just reserved for future usage. I did similar things for download files. 
If autoSave is turned on, frontend/browser will save the blob. It can be used when clicking a link/button then download/save starts.
But for rendering the bytes on browser we just need the blob; hence autoSave should be set to false.

lmarini (approved):

PR #577, opened 2023-07-06
change models for visualizations

tcnichol:
This pull request changes models for visualizations based on feedback.


longshuicy (approved):

backend/app/models/visualization_config.py
```diff
@@ -16,8 +16,8 @@ class VisualizationConfigBase(BaseModel):
     extractor_info: Optional[ExtractorInfo]
     job: Optional[EventListenerJobDB]
     client: Optional[str]
-    viz_config_data: dict = {}
-    # TODO add json document or key value pairs, config_variables
+    parameters: dict = {}
+    visualization_bytes_id: List[PyObjectId]
```
longshuicy, line 20: maybe plural  `_ids` since it's a list

backend/app/models/visualization.py
```diff
@@ -24,21 +24,21 @@ def validate_to_json(cls, value):
...
+class VisualizationDataDB(Document, VisualizationDataBase):
     creator: UserOut
     created: datetime = Field(default_factory=datetime.utcnow)
     modified: datetime = Field(default_factory=datetime.utcnow)
     bytes: int = 0
     content_type: ContentType = ContentType()
 
     class Settings:
-        name = "visualizations"
+        name = "visualizationData"
```
lmarini, line 39: Do you we use camel case anywhere? Should we use all lower case with underscore between each word instead?

lmarini (changes requested):

ddey2:
@tcnichol 
Should we change "vizconfig" in 29 in isualization_config.py to "visualization_config" may be?

backend/app/models/visualization.py
tcnichol: I think that will be better. Changed.

longshuicy:
Could you resolve merge conflict?

longshuicy (approved):

max-zilla (approved):

backend/app/routers/visualization.py
tcnichol, line 1: The credentials are removed. They were causing this route to give a 403 error.

PR #580, opened 2023-07-07
Updating visualization download url

ddey2:


longshuicy (approved):

lmarini (changes requested):
My original comment was to change this to `/visualizations/${visualizationId}/bytes` instead of `/download/${visualizationId}/bytes`. This would be inline with RESTful web service design.

https://github.com/clowder-framework/clowder2/pull/573#discussion_r1253661495

sandeep-ps (approved):
Looks good to me. I'll approve as well.

PR #593, opened 2023-07-13
visualization_config_id now in the visualization_data object

tcnichol:
I think this fixes the issue, though was not sure if routes needed to be changed or if that would all happen in previews/extractors or pyclowder.


max-zilla (changes requested):
I think this should also include the functionality that hitting DELETE endpoint for visualization_config should delete any vis_data associated with that config.

tcnichol:
> I think this should also include the functionality that hitting DELETE endpoint for visualization_config should delete any vis_data associated with that config.

This should be the case now both ways.

ddey2 (approved):
looks good.

longshuicy (approved):
Works well. 
I only have a question about vis config attaching the whole vis data model as its output. 
Would it be too much of the payload? 
@lmarini @max-zilla thoughts?

max-zilla (approved):

PR #608, opened 2023-07-21
fix adding non-existent user to group

tcnichol:
This fixes the message when trying to add a user to a group that does not exist. Same error message as when adding non-existent user elsewhere.


frontend/src/components/datasets/ShareDatasetModal.tsx
```diff
@@ -74,6 +74,7 @@ export default function ShareDatasetModal(props: ShareDatasetModalProps) {
 
 	const onShare = async () => {
 		await setUserRole(datasetId, email, role);
+		console.log('clicked on share',datasetId, email, role);
```
longshuicy, line 77: can you remove the console.log?

tcnichol: fixed

longshuicy:
The title of this PR is confusing since sharing dataset with not existing user still works (which i think is a bug). 
But you fixed adding non existing user to group.

longshuicy (approved):

PR #609, opened 2023-07-21
fix logout

tcnichol:
logout was failing since token_exist is now an object, not a dict. So when it tried to find token_exist["refresh_token"] nothing was there.


longshuicy (approved):
works well now. Good job identifying the issue.

PR #611, opened 2023-07-25
597 edit group name and description

tcnichol:
We can now edit the name of a group.

Additionally, the edit_group route should now handle all cases of adding users, removing users, or changing editor value for users.

### DIFF
--- a/backend/app/routers/groups.py
+++ b/backend/app/routers/groups.py
@@ -121,7 +121,9 @@ async def edit_group(
         for original_user in original_users:
             if original_user not in groups_users:
                 # remove them from auth
-                async for auth in AuthorizationDB.find({"group_ids": ObjectId(group_id)}):
+                async for auth in AuthorizationDB.find(
+                    {"group_ids": ObjectId(group_id)}
+                ):
                     auth.user_ids.remove(original_user.user.email)
                     await auth.replace()
                     # Update group itself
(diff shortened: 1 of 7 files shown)


backend/app/routers/groups.py
```diff
@@ -104,20 +104,56 @@ async def edit_group(
...
+                await group.replace()
+                # Add user to all affected Authorization entries
+                await AuthorizationDB.find(
+                    AuthorizationDB.group_ids == ObjectId(group_id),
+                ).update(
+                    Push({AuthorizationDB.user_ids: user.email}),
+                )
         try:
-            group.update(group_dict)
+            group.name = group_dict["name"]
```
longshuicy, line 156: hmm can user also update the description? I only see you overwrite the name part

longshuicy (approved):

frontend/src/components/groups/EditNameModal.tsx
```diff
@@ -0,0 +1,87 @@
...
+	};
+
+	const handleDialogClose = () => {
+		handleClose();
+		dispatch(resetFailedReasonInline());
+	};
+
+	const onSave = async () => {
+		setLoading(true);
+		console.log(groupId, groupName,"oldname", name,'newname')
```
longshuicy, line 54: could you remove the console.log?

backend/app/routers/groups.py
tcnichol: I can add an update description.

frontend/src/components/groups/EditNameModal.tsx
tcnichol: removed

lmarini:
pytest failing

max-zilla (approved):

PR #612, opened 2023-07-25
596 prettify the multi grid layout of the visualization

longshuicy:
- Wrap each visualization in card and break them down to smaller custom components
- Add toggle view to display data information or config paramters

Pasted Graphic 8

Technical Dive into


tcnichol (approved):
Could not test Geospatial but was able to test everything else. I will mark approved.
@max-zilla if you have had Geospatial work with this branch, then it should be okay.

max-zilla (approved):

PR #614, opened 2023-07-26
Fixed submission to extractors in prod

lmarini:
One of the methods was using hardcoded rabbitmq credentials and URL.

Renamed docker service to traefik to be more explicit.

Renade route to post from get.

### DIFF
--- a/backend/app/rabbitmq/listeners.py
+++ b/backend/app/rabbitmq/listeners.py
@@ -59,7 +59,6 @@ async def submit_file_job(
     token: str = Depends(get_token),
 ):
     # Create an entry in job history with unique ID
-    print("IN_submit_file_job")
     job = EventListenerJobDB(
         listener_id=routing_key,
         creator=user,
--- a/backend/app/rabbitmq/listeners.py
+++ b/backend/app/rabbitmq/listeners.py
@@ -23,10 +23,7 @@ from app.routers.users import get_user_job_key
 
 
 async def create_reply_queue():
-    credentials = pika.PlainCredentials("guest", "guest")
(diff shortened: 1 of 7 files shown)


max-zilla (approved):

docker-compose.yml
```diff
@@ -88,30 +91,30 @@ services:
...
     build:
-      context: ./backend
+      context: backend
       dockerfile: messages.Dockerfile
     environment:
-      - MONGODB_URL=mongodb://mongo:27017
-      - RABBITMQ_HOST=rabbitmq
-      - RABBITMQ_USER=${RABBITMQ_USER:-guest}
-      - RABBITMQ_PASS=${RABBITMQ_PASS:-guest}
+      MONGODB_URL: mongodb://mongo:27017
```
ddey2, line 114: Just a query..... I noticed there was a difference in pattern there env were written. I guess it didn't make any changes to the workflow apart from aesthetics, right?

lmarini: one is a map and the other a list in yml. I was trying to see it would make a difference with the interpolation, but I think both seem to work fine.

backend/app/rabbitmq/listeners.py
```diff
@@ -62,6 +59,7 @@ async def submit_file_job(
     token: str = Depends(get_token),
 ):
     # Create an entry in job history with unique ID
+    print("IN_submit_file_job")
```
ddey2, line 62: Do we need these print statements?

ddey2 (approved):

backend/app/rabbitmq/listeners.py
lmarini: no, thank you

PR #616, opened 2023-07-27
Moving createFileMetadata so that it's called after a file update is successful

ddey2:
Steps to reproduce

1. Try to upload a file
2. Try to update that file with another file with different name or content type

Expected behavior - It should show correct error message. 
Screenshot 2023-07-27 at 10 59 42 AM


tcnichol (approved):
Tested this one with name and type, both show correct error message.

max-zilla (approved):

Issue #541, opened 2023-06-08
Selecting both start and end date search filters crashes search page


max-zilla:
<img width="1322" alt="image" src="https://github.com/clowder-framework/clowder2/assets/14166971/fa149d42-ebeb-4c4c-9281-42d63c3a64a0">
When just one field is populated, it is OK. when both are selected, the page crashes.
<img width="1368" alt="image" src="https://github.com/clowder-framework/clowder2/assets/14166971/3502a07a-20c4-45d1-8b86-a660ba774726">

max-zilla:
PR opened: https://github.com/clowder-framework/clowder2/pull/619

PR #619, opened 2023-07-28
replace DateRange with two DatePickers

max-zilla:
Issue #541 fix
This fixes a bug with reactivesearch's DateRange field that was causing infinite recursion trying to update the range according to the results returned from itself.

Implementing two individual DatePicker fields that have custom queries for gte/lte (Start/End Date) resolves this issue, seems to work as desired now.


tcnichol (approved):
Looks good. Search returns results from within the range.

longshuicy (approved):

Issue #603, opened 2023-07-19
Multiple metadata fields when creating dataset result in multiple entries in search results


lmarini:
I had 4 required fields when creating a dataset. After filling them out and creating the dataset, searching for the name of the datasets shows four results in the search result list (I am guessing one for each metadata entry).

PR #626, opened 2023-08-04
change ESMetadata field names

max-zilla:
This is for #603 

THIS PR OVERHAULS ELASTICSEARCH INDEX. Recommend stopping your containers, deleting volumes, restarting, then creating a Metadata Definition and creating datasets/files. This should resolve duplicate results bug and re-enable Lucene search. 

It was becoming difficult to make effective search while files/datasets/metadata were in separate Elasticsearch indexes, here they are merged into one ElasticsearchObject closer to v1 that includes metadata directly on object. 

Examples:

File and dataset with metadata:
image
image

Use fancy lucene syntax:
image

Combine them with another dataset that has Orange as color:
image

Multiple criteria:
image
image


lmarini (changes requested):
I wasn't able to get this to work. I did drop the indexes. Documents show in elasticsearch but the UI returns "no results". I created a dataset with name/description of "foobar" and two metadata fields, one a text one with content of "foobar". Searching for "foobar" doesn't return documents. Searching with empty string also doesn't return documents.

tcnichol:
With advanced search, I see datasets by name and from metadata entries, started with new db so didn't need to re-index.

And I don't see results if I don't have permission for a particular user.

Is there anything else that needs to be tested? That seems like everything.

tcnichol (approved):
Tested with new datasets and metadata.
No duplicates, correct datasets returned, only datasets with permission returned. 
Approved.

longshuicy:
what's the proper way to drop indexes? I delete the whole volume but still see below behavior
- when turn on lucene i see double results 
- I can't search any metadata (file and dataset names works fine). I'm also missing the creator dropdowns

image

image

longshuicy (approved):
Approved. 
No duplicate results, and lucene search works with metadata.

ddey2 (approved):
Looks good.

PR #630, opened 2023-08-07
add messages extractor to github build so it can be pushed to docker.io

longshuicy:
In order to retrieve message image from docker.io, adding this github flow


lmarini:
Could we build and push messages and heartbeats together with the backend in one action? with maybe 3 docker/build-push-action@v3 blocks? I it should speed up build process and it makes sense to have all of them pass or fail together?

lmarini (approved):

PR #632, opened 2023-08-07
fix dataset submit after metadata

tcnichol:
Before you could not submit even after you put in values for required metadata.
The problem was using .every instead of .map. With .map finish button is active once all fields are filled out.


longshuicy (changes requested):
There is another place that uses every. Could you fix that if that's not too much? 
image

tcnichol:
> There is another place that uses every. Could you fix that if that's not too much? image

Changed this.

longshuicy (approved):

PR #633, opened 2023-08-08
change file version

tcnichol:
Right now we can change the file version. 

The file summary has not changed to reflect the old version information yet. That will be done after metadata. Only the name of the left component changes.

I am now sending the selected version to the file metadata, but I am not sure that this is providing the correct version's metadata. Will need to start over with a new file with new metadata to test. 

Another thing that does not work. If you have the metadata tab open, and then you change the file version using the modal, it does not change to reflect the new file version. I was not sure how to handle this.


tcnichol:
The following now works:
Changing file version will change the ListenerMetadata and the File Details is replaced with File History, where bytes, created, author are replaced with values from the previous version. 

Right now there is no change for visualization. Will make that a separate issue.

To test, upload a text file, run wordcount extractor, and then modify it to create some new versions and re-run extractor. When you change, you will see the metadata change and also, the file details will change.

max-zilla (changes requested):
There is a bug where I uploaded Squirrel.gif, then tried to upload a new file version using OtherFile.gif. The API does not allow a new file version to have different filename from before, so it returns a 400, but the UI does not seem to catch this and tries to show the newer version of the file (that didn't successfully get created). Not sure if this bug was introduced here but we should fix it here. (see routers/files `update_file` method for this logic - we should catch the 400)
image

tcnichol:
Debugging with the file update, I'm seeing now that the right error message does pop up first:
correct_error
Then it looks like the wrong one comes after that. Trying to figure out what is going on.

max-zilla (approved):
tested uploading a file and triggering a rejection on the update, switching versions, everything seemed to work.

max-zilla:
@tcnichol i think if you resolve merge conflict we can merge

lmarini:
To do in this or future PR:
- make button a pull down
- remove change version buttons (next to each version)

ddey2 (approved):

tcnichol:
Now the file version should be next to and not underneath the filename in the files table. 
Screenshot 2023-08-24 at 3 58 37 PM

PR #634, opened 2023-08-10
breadcrumb back for dataset and folder and subfolder

tcnichol:
The breadcrumb is now back for datasets, folders, and subfolders.

### DIFF
--- a/frontend/src/components/datasets/Dataset.tsx
+++ b/frontend/src/components/datasets/Dataset.tsx
@@ -190,10 +190,10 @@ export const Dataset = (): JSX.Element => {
 				{/*title*/}
 				<Grid item xs={8} sx={{ display: "flex", alignItems: "center" }}>
 					<Stack>
-						<Box>
-							<MainBreadcrumbs paths={paths}>
-							</MainBreadcrumbs>
-						</Box>
+						{/*<Box>*/}
+						{/*	<MainBreadcrumbs paths={paths}>*/}
+						{/*	</MainBreadcrumbs>*/}
+						{/*</Box>*/}
 						<Box
 							sx={{
 								display: "inline-flex",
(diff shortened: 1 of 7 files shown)


frontend/src/components/datasets/Dataset.tsx
```diff
@@ -49,7 +50,7 @@ export const Dataset = (): JSX.Element => {
 	// search parameters
 	const [searchParams] = useSearchParams();
 	const folderId = searchParams.get("folder");
-
+	console.log('folder id is', folderId);
```
longshuicy, line 53: could you remove the console log?

frontend/src/components/datasets/Dataset.tsx
```diff
@@ -280,6 +281,13 @@ export const Dataset = (): JSX.Element => {
 							disabled={false}
 						/>
 					</Tabs>
+					{folderId !== null ?
+							<Box>
```
longshuicy, line 285: Would aligning with the file table will make it look cleaner? Thoughts?

image


Add 
```
 padding: "24px 24px 0 24px"
```

longshuicy (changes requested):

frontend/src/components/datasets/Dataset.tsx
tcnichol: @longshuicy 

I added this but somehow it's not aligned properly, but it looks like it's okay on the Files component. Not sure what I did wrong.

frontend/src/components/datasets/Dataset.tsx
tcnichol: removed

frontend/src/components/files/File.tsx
```diff
@@ -308,6 +307,7 @@ export const File = (): JSX.Element => {
 							disabled={false}
 						/>
 					</Tabs>
+					<MainBreadcrumbs paths={paths} />
```
longshuicy, line 310: you need to add the same thing here 
```
<Box style={{padding: "24px 24px 0 24px"}}>
```

longshuicy (approved):
Approves.

max-zilla (approved):
looks good.

PR #637, opened 2023-08-11
improve the logic of handling incomplete visualization configuration

longshuicy:
consider the case that parameters are also empty.


tcnichol (approved):
This looks good, works in the case that these fields are empty.

longshuicy:
For some reason i can't get the pytest to work...

PR #640, opened 2023-08-16
Fix to dataset extractor

ddey2:
fixes to dataset.py to make dataset extractor working.

To test, please run test-dataset-extractor from https://github.com/clowder-framework/pyclowder/pull/67

Trigger it on any dataset.


longshuicy (approved):

tcnichol:
Yes.

There was a dependency for Security that we no longer use that only works with token, but not API key. It was removed in other routes but was still in the file upload. 

Token was used in methods but, in the end, wasn't used in the last one, so I removed it. The user info is still used with rabbitmq and posting the jobs.

tcnichol (approved):
Tested.

Works with the pyclowder branch, the file posts into the dataset now. Approved.

tcnichol:
I made 1 fix here based on the debugging from Friday.

Mimetypes.get_content_type was being given a io.FileBytesIO, but it needed the file name. So it now is sent the file name from the new_file File object that is created.

I tested and this still works with visualization. 

Since I had it working I went ahead and pushed here while I remembered. 

With the dataset extractor, the file that gets sent does have None for a filetype but this does correctly find the filetype from the filename and the entry is added in minio. When downloaded the file is the same now.

ddey2:
@tcnichol Thanks. It needed one little change. Now everything is working.

PR #646, opened 2023-08-19
Edit Group Name no longer erases group description

tcnichol:
When editing the name now the group Description stays the same. The group description was not being sent to that modal and was not sent as formData to the backend. Now it is.


longshuicy (approved):
This fixes the issue. Looking good approve.

max-zilla (approved):

PR #652, opened 2023-08-21
Backend now pickups JWT token in cookie left by frontend and

lmarini:
uses it to authenticate users. This should let us open a download window for files and datasets that will not run out of memory unlike the download in javascript approach.

### DIFF
--- a/backend/app/keycloak_auth.py
+++ b/backend/app/keycloak_auth.py
@@ -54,8 +54,8 @@ jwt_header = APIKeyCookie(name="Authorization", auto_error=False)
 
 
 async def get_token(
-        token: str = Security(oauth2_scheme),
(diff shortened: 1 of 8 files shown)


tcnichol:
I'm having an issue uploading a large file on this branch. It's a 39 GB zip archive. I see this:
upload_error

ddey2:
I tried with 1.5GB file. It works fine, within few seconds. I couldn't see a separate tab opened when I downloaded the file.

longshuicy:
@tcnichol @ddey2 The frontend still need to change how upload/download works to address the large file issue. I will address them in another PR. This PR provides the backend support for the change.

longshuicy:
Everything should work with download now

PR #653, opened 2023-08-21
651 do not use extractor info

tcnichol:
Pull request. Dependent on this pyclowder pull request

https://github.com/clowder-framework/pyclowder/pull/81

I also took out a few uses of the access token from the now unused Security dependency in some extractor related files.


tcnichol:
This branch should fix the problem of the listener metadata not showing up. Most recent changed from main have been merged in.

lmarini (approved):

PR #657, opened 2023-08-23
Rewrite upload file

longshuicy:
Rewrite the upload using plain input/file
Tested with 1.2G data locally and it seems to be ok
image


ddey2 (approved):
Tried uploading 1.1GB. It worked comparatively faster.

max-zilla (approved):

PR #658, opened 2023-08-24
add increment flag to download endpoint

max-zilla:
This adds a flag for frontend to use on file download endpoint that disables incrementing of download count, so refreshing page wont increase # of downloads if the visualization downloads the bytes for any reason.


longshuicy (approved):
Tested with both download and visualization raw bytes, work as expected. Approve.

ddey2 (approved):

PR #666, opened 2023-08-24
adding file size threshold limit on client side

ddey2:
Initial implementation : I have put the logic on client side to check for file threshold for visualization.

I am now looking into ways to put the entire logic in backend.

To test: 
1. Upload a file greater than threshold size
2. It should not show the file in visualization tab and show a console log.

### DIFF
--- a/frontend/src/app.config.ts
+++ b/frontend/src/app.config.ts
@@ -15,6 +15,7 @@ interface Config {
 	extractorInterval: number;
 	eventListenerJobStatus: EventListenerJobStatus;
 	streamingBytes: number;
+	rawDataVisualizationThreshold: number;
 }
 
 const config: Config = <Config>{};
@@ -58,5 +59,6 @@ config["eventListenerJobStatus"]["skipped"] = "SKIPPED";
 config["eventListenerJobStatus"]["resubmitted"] = "RESUBMITTED";
 
 config["streamingBytes"] = 1024 * 10; // 10 MB?
+config["rawDataVisualizationThreshold"] = 1024 * 1024 * 10; // 10 MB
 
 export default config;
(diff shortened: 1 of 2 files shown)


frontend/src/components/visualizations/Visualization.tsx
```diff
@@ -94,10 +95,14 @@ export const Visualization = (props: previewProps) => {
 								// if no visualization config exist, guess which widget to use by looking at the mime type of
 								// the raw bytes
 								else {
+									//check if file size is greater than threshold, if it is, then show no visualization
+									if (fileSummary && fileSummary.bytes && fileSummary.bytes >= config["rawDataVisualizationThreshold"]) {
+										console.log("File is greater than threshold");
```
longshuicy, line 100: Short term i think this works. Long term we would benefit from thinking about backend solution to handle this logic.
Also mayb ewe could have a component that display "file is too big for visualization"?

frontend/src/app.config.ts
```diff
@@ -58,5 +59,6 @@ config["eventListenerJobStatus"]["skipped"] = "SKIPPED";
 config["eventListenerJobStatus"]["resubmitted"] = "RESUBMITTED";
 
 config["streamingBytes"] = 1024 * 10; // 10 MB?
+config["rawDataVisualizationThreshold"] = 1024 * 1024 * 10; // 10 MB
```
longshuicy, line 62: You might be able to just use the "streamingBytes" ^. I remembered i might reserved that but never used it

longshuicy (approved):

frontend/src/app.config.ts
ddey2: I didn't realize you don't use streamingBytes.

PR #668, opened 2023-08-25
663 frontend download button onclick update the summary immediately

longshuicy:
Point this to the new download PR.
To test, download file and watch the download counts immediate change

image


ddey2:
I see a some weird thing. I click on 'Download'. It showed 'some backend error' on UI (though I couldn't find anything in pycharm). Then the download count is not incremented immediately. On page refresh, it is incremented twice (the count has gone from 5 to 7). Can somebody else check?
Screenshot 2023-08-25 at 10 22 09 AM

Screenshot 2023-08-25 at 10 21 56 AM

max-zilla (approved):

PR #669, opened 2023-08-25
200 fix hardcoding pieces in helm chart

longshuicy:
Load the keycloak realm in as ConfigMap. You can test by deleting the Realm in keycloak and then restart the keycloak pods see if it gets loaded correctly.

### DIFF
--- a/deployments/kubernetes/charts/clowder2/files/realm.json
+++ b/deployments/kubernetes/charts/clowder2/files/realm.json
@@ -1,0 +1,1914 @@
+{
+  "id": "b5eacc0b-a10a-49b7-8f21-ab93c71735a9",
+  "realm": "clowder",
+  "notBefore": 0,
+  "defaultSignatureAlgorithm": "RS256",
+  "revokeRefreshToken": false,
+  "refreshTokenMaxReuse": 0,
+  "accessTokenLifespan": 300,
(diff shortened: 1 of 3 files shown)


robkooper (changes requested):
This will hard-code the realm to be for a specific instance, and will overwrite the changes that the user makes on a restart of keycloak.

deployments/kubernetes/charts/clowder2/files/realm.json
robkooper, line 1: This is a massive file, and not sure all the parameters are needed if they are the defaults.

longshuicy: there is a mini version of that but i haven't test it

longshuicy:
> This will hard-code the realm to be for a specific instance, and will overwrite the changes that the user makes on a restart of keycloak.

Ah right... what's a better way then, using initContainers?

robkooper:
I would have it as an optional hook (post-install) in the helm chart when you install it (not upgrade). I would have a boolean flag that is "install realm" and have it be false. This way the user can decide that you want to import it.

Problem with init is that if you are not careful it will overwrite any changes the user has made.

longshuicy:
Created a new task for it #689 for the future.

lmarini (approved):

PR #672, opened 2023-08-25
`submit_file_job` takes 5 parameters, put we had 6. 

lmarini:
The extra parameter `access_token` was throwing an error when submitting a file for extraction.


tcnichol (approved):
This fixes the error. Was able to submit. 

Approved.

PR #674, opened 2023-08-25
671 file breadcrumbs in the wrong spot

tcnichol:
I think this is fixed. Luigi, let me know if this looks right.


longshuicy:
is it intended that for file it shows at the top, but for folder it shows under the tabs? 
image
vs
image

tcnichol:
@longshuicy that was what Luigi wanted, but @lmarini can you check this just to be sure? I also think that changing from button to link removed any white space before the folder name.

tcnichol:
I have a few more changes to make here so I changed this to draft.

lmarini (approved):

ddey2 (approved):

PR #681, opened 2023-08-28
Update populate fake data

lmarini:
Add csv files to fake datasets.
Create random password for users.
First user becomes ownder of every dataset.


tcnichol:
Luigi, 

I checked this one out and it throws an error because Pillow isn't in the dependencies. But when I did a pip install Pillow it works.

longshuicy:
There is a failed github action (which shouldn't be there since it suppose to only check "backend" folder). Opened a task #703 

Other than that i have been successfully run the script. But since those datasets are created by different users, any idea i can actually see them as "myself". Maybe add a step to "share" or making them public?

max-zilla:
In my case, it looks like different people create the datasets but they are all shared with the first user so the first user can see them. Logging in as the different users will show/hide different things.

longshuicy:
> In my case, it looks like different people create the datasets but they are all shared with the first user so the first user can see them. Logging in as the different users will show/hide different things.

Ah sorry my fault. I didn't read the script print out carefully.

tcnichol:
@lmarini i have this pull request open that I wonder if it will fix the github action errors here:

https://github.com/clowder-framework/clowder2/pull/706

lmarini:
@tcnichol @longshuicy I have run black on the scripts folder and it fixed the user. I think it might be a good thing to make sure the python code in the scripts folder is also formatted properly. Should we close the other branch that tries to make the action run only in backend? Or do we want to keep it and add `scripts` to the list? Maybe it should just run on the all project like it currently does?

tcnichol:
@lmarini 

Whole project is probably best, unless you can think of any reason why any particular code shouldn't be formatted. So you can go ahead and close that branch.

tcnichol (approved):
All tests pass. I can sign in as fake users using the fake user passwords and view the data and see that it is shared.

max-zilla (approved):

PR #682, opened 2023-08-29
fix thumbnail display, update search result icons

max-zilla:
This uses the new thumbnail endpoints to display thumbnails for files in a dataset.

### DIFF
--- a/frontend/src/components/files/FilesTableFileEntry.tsx
+++ b/frontend/src/components/files/FilesTableFileEntry.tsx
@@ -9,7 +9,7 @@ import { parseDate } from "../../utils/common";
 import FileMenu from "./FileMenu";
 import prettyBytes from "pretty-bytes";
 import { FileOut } from "../../openapi/v2";
-import { generateVisDataDownloadUrl } from "../../utils/visualization";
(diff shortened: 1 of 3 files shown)


frontend/src/components/files/FilesTableFileEntry.tsx
```diff
@@ -27,7 +27,7 @@ export function FilesTableFileEntry(props: FilesTableFileEntryProps) {
 			try {
 				let url = "";
 				if (file.thumbnail_id) {
-					url = await generateVisDataDownloadUrl(file.thumbnail_id);
+					url = await downloadThumbnail(file.thumbnail_id);
```
longshuicy, line 30: Since now we enabled cookie authentication, we might be able to just say
```
<img src={`${config.hostname}/api/v2/thumbnails/${file.thumbnailId}`} alt="thumbnail" width="24" />
```

tcnichol:
I am using the dataset test extractor but the thumbnail on the dataset is still coming up broken for me. Is there a different way to test this? Not sure I was doing it right.
Screenshot 2023-08-28 at 1 04 53 PM

longshuicy (approved):
Will address my comment in a seperate PR. This looks good.

PR #684, opened 2023-08-29
660 authentication timeout on search page

longshuicy:
The root cause is the updated cookie is not pulled correctly. I tried so many different things (hook, redux/reducer, etc) but none works except using time interval to periodically pull the cookie. 

I also revived the topbar embedded search with same mechanism.

To test:
1. Sit on any of the non-search page for > 5 min and then try search on the top bar see if there is result prompted 
2. Sit on any of the non-search page for > 5 min and then go to search page see if the facet got populated correctly
3. Sit on the search page for > 5 min and search

### DIFF
--- a/frontend/src/actions/authorization.js
+++ b/frontend/src/actions/authorization.js
@@ -1,11 +1,14 @@
-import {V2} from "../openapi";
-import {handleErrorsAuthorization, resetFailedReason} from "./common";
+import { V2 } from "../openapi";
+import { handleErrorsAuthorization, resetFailedReason } from "./common";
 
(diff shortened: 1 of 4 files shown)


lmarini (approved):
@longshuicy this PR looks like it's working fine. The only thing is that every minute you sit on the search page it auto refreshes.  I am assuming because the token is being refreshed. Let's merge it and make a new issue to see if we can fix it later on or until we rebuild the search interface from scratch?

max-zilla (approved):

max-zilla:
issue: https://github.com/clowder-framework/clowder2/issues/694

PR #686, opened 2023-08-30
623 geoserver and geotiff previewer helm chart

longshuicy:
Geoserver deployed 
Geotiff previewer also deployed
Almost ready to be reviewed


max-zilla (approved):
tested on the kube deployment, found separate issue #716 but otherwise good.

PR #688, opened 2023-08-30
drop test database and test users before running tests

tcnichol:
We now delete both the test users and the existing test database before running tests. This eliminates errors that happen when the test user already exists in keycloak but the tests try to create the user again.


backend/app/tests/conftest.py
```diff
@@ -23,6 +23,7 @@ def root_path() -> str:
 
 @pytest.fixture(scope="session")
 def token(client: TestClient) -> str:
+    print("do this first")
```
lmarini, line 26: can this be removed or updated?

lmarini (changes requested):

backend/app/tests/conftest.py
tcnichol: removed, not needed.

PR #690, opened 2023-09-01
Docker compose cleanup

lmarini:
- Removed restart: unless-stopped from docker compose dev
- Removed old docker compose files.


tcnichol:
I ran the docker-compose files (dev and docker compose) and see all the volumes created and everything works. Anything specific that I should look at for testing beyond that? If that is all I can mark approved.

tcnichol:
On this branch, I am noticing that the container for the extractor heartbeat stops at some point and is no longer running. It does start initially but by the time I ran some extractors, the container was gone.

max-zilla:
@tcnichol i think this is because RMQ takes a minute or two to be ready, but the messages and heartbeat extractors start up immediately and can't connect. I think this is originally why I added the restart flag for those. I think the solution is either to kill and let docker restart the messages/heartbeat containers after a couple minutes when RMQ is up, or we modify those services to stay in loop trying to reconnect if it fails.

PR #692, opened 2023-09-01
685 retire the function that generate frontend url for visualization

longshuicy:
Update all the "download url" related method and usage to directly use the /api endpoint since now cookie auth is supported

Things to test:
1. Dataset Thumbnail
2. File thumbnail
3. Image vis
4. Video vis
5. Audio vis
6. Iframe


max-zilla (approved):
the various thumbnails and visualizations i had all seem to work as expected.

PR #700, opened 2023-09-06
697 folder doesnt not aligned with the rest of the files

longshuicy:
image
image

Fix both cases


tcnichol:
Looks good.

Another possible issue. For folders, 'size' is blank. Would it make sense to have that be combined size of all the files inside a folder? 

Though that's another number we'd have to keep updating or compute every time.

tcnichol (approved):
The right content is in the right columns here. Marking approved.

lmarini:
> Looks good.
> 
> Another possible issue. For folders, 'size' is blank. Would it make sense to have that be combined size of all the files inside a folder?
> 
> Though that's another number we'd have to keep updating or compute every time.

That's sounds like a great idea @tcnichol Want to open an issue for it? Not sure we can work on it now, but it would be awesome to do later on.

PR #705, opened 2023-09-06
695 clean up the primary and secondary styling

longshuicy:
Couldn't figure out when to use the secondary color. So for now I tried to stick to primary with various shade, or black.

image
image


longshuicy:
Pointing this branch to the @lmarini  your `mui-colors` branch

lmarini (approved):
Looks great, thank you. Next time feel free to push changes to the other branch.

Just realized that the login page is in keycloak. How easy would it be to update that?

PR #706, opened 2023-09-06
specify backend for running black formatting

tcnichol:
I am not sure if this is right, but this looks like it might fix the issue on Luigi's populate fake data pull request.


longshuicy (approved):

ddey2 (approved):

tcnichol:
I think we decided it was okay to have formatting check in scripts, so I am closing this one.

PR #708, opened 2023-09-06
Add ability to create Extractors 100% in the GUI

KastanDay:
GUI Extractor Requirements:

> These are living notes, drawing from inspiration

1. Launch-able in GUI.
    - Implementation details
        - CPU-only for now
        - Run on your Clowder instance
        - Unsafe code is okay, just have a big red warning.
        
2. Can connect to them just to hit endpoints with curl. 

```bash
curl -X POST \
    -H "Content-Type: application/json" \
    -d '{"prompt": "Software developers start with a Hello, [MASK]! script."}' \
    http://localhost:8000/
```

3. Can connect via PyClowder SDK

```python
from banana_dev import Client

# Create a reference to your model on Banana
my_model = Client(
    api_key="YOUR_API_KEY", # Found in dashboard
    model_key="YOUR_MODEL_KEY", # Found in model view in dashboard
    url="http://localhost:8000",
)

# Specify the model's input JSON
inputs = {
    "prompt": "Software developers start with a Hello, [MASK]! script",
}

# Call your model's inference endpoint
result, meta = my_model.call("/", inputs)
print(result)

# result:
# {
#     "outputs": {
#         "score": 0.8281897306442261,
#         "sequence": "software developers start with a hello, world! script.",
#         "token": 2088,
#         "token_str": "world"
#     }
# }
```

4. Custom preprocessing

```python
# https://github.com/replicate/cog
from cog import BasePredictor, Input, Path
import torch

class Predictor(BasePredictor):
    def setup(self):
        """Load the model into memory to make running multiple predictions efficient"""
        self.model = torch.load("./weights.pth")

    # The arguments and types the model takes as input
    def predict(self,
          image: Path = Input(description="Grayscale input image")
    ) -> Path:
        """Run a single prediction on the model"""
        processed_image = preprocess(image)
        output = self.model(processed_image)
        return postprocess(output)
```

1. Arbitrary code execution. 
    1. Limit outbound requests from container.

CleanShot 2023-09-06 at 13 47 10

### DIFF
--- a/frontend/src/components/listeners/CreateListener.tsx
+++ b/frontend/src/components/listeners/CreateListener.tsx
@@ -1,0 +1,366 @@
+import React, { useEffect, useState } from "react";
+
+import {
+	Autocomplete,
+	Box,
+	Button,
+	ButtonGroup,
+	Divider,
+	FormControl,
+	FormControlLabel,
+	FormHelperText,
+	FormLabel,
+	Grid,
+	IconButton,
+	InputBase,
+	List,
+	Radio,
+	RadioGroup,
+	TextField,
+} from "@mui/material";
+import { useDispatch, useSelector } from "react-redux";
+// import { RootState } from "../../types/data";
+
+// import { CreateListenerModal } from "./CreateListenerModal";
+// import { CreateMetadata } from "../metadata/CreateMetadata";
+// import {
+// 	fetchMetadataDefinitions,
+// 	postDatasetMetadata,
+// } from "../../actions/metadata";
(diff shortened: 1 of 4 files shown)


frontend/src/components/listeners/CreateListener.tsx
KastanDay, line 1: dw, I'll clean up these comments. Just keeping them for now as I do the initial implementation.

KastanDay:
Maybe start with SegmentAnything model for Xiaoxia Liao to use.

PR #717, opened 2023-09-07
679 download button text disappears at small widths

tcnichol:
The download button is now visible with the full word 'download' inside of the button. Even with longer dataset names this seems to work.


ddey2 (approved):
works with resizing window and very long dataset name

lmarini:
@tcnichol  I have reverted the title grid cell size to 8 and 4. It was leaving toomuch white space between title on the left and buttons on the right. Not sure if there was another reason to change that.

PR #724, opened 2023-09-11
Adding a card for no content

ddey2:


ddey2:
Screenshot 2023-09-11 at 5 09 51 PM
 Somehow it's rendering it 6times. I am looking into why/where this piece of code is getting called.

frontend/src/components/visualizations/Visualization.tsx
```diff
@@ -87,6 +88,9 @@ export const Visualization = (props: previewProps) => {
 														"No visualization data or parameters available. " +
 															"Incomplete visualization configuration."
 													);
+													return (
+											<NoVisualizationCard msg = "No visualization data or parameters available. Incomplete visualization configuration."></NoVisualizationCard>
```
longshuicy, line 92: The reason why cards showing so many times is because this is a loop that check for any registered visualization configuration/ or check for any matching file content type. Some times there are many applicable match.

I would suggest instead create two states to indicate "noVis" and "fileOverThreshold" flag. Then whenever you run into the situation, just turn on those flags.
Have a use Effect to listen to those flag change. If flag is True display the card accordingly.

longshuicy (approved):
Functionality works. We might want to improve the visual in a subsequent task?

ddey2:
Updated the ui. I think for now this should be good. We can alter it later if needed. 
Screenshot 2023-09-13 at 12 24 06 PM

max-zilla:
Generally works as expected. However, if image files are small enough to use raw bytes but there is no viz config, it still shows the error message along with the raw preview which might be confusing: 
image

I think we should only show this message if we can't do anything with the raw bytes. Not sure how easy that is to determine right now though (i.e. would we have to check MIME types and stuff?)

frontend/src/components/visualizations/Visualization.tsx
```diff
@@ -95,21 +117,16 @@ export const Visualization = (props: previewProps) => {
...
-										fileSummary &&
+									//to make sure file size is less than threshold
+									if (
+										fileSummary && fileSummary.bytes && fileSummary.bytes < config["rawDataVisualizationThreshold"] &&
 										fileSummary.content_type !== undefined &&
 										((fileSummary.content_type.content_type !== undefined &&
 											visComponentDefinition.mimeTypes.includes(
 												fileSummary.content_type.content_type
 											)) ||
-											(fileSummary.content_type.content_type === undefined) && (fileSummary.content_type.main_type !== undefined &&
```
max-zilla, line 112: use this logic the same way in the useEffect above to check all the visComponentDefinitions for a matching mime type - only if we don't have a component to show the raw bytes show the "No visu"

frontend/src/components/visualizations/Visualization.tsx
```diff
@@ -42,8 +46,31 @@ export const Visualization = (props: previewProps) => {
 		}
 	}, [fileId, datasetId]);
 
+	// Check for conditions and set state only once
+	useEffect(() => {
+		if (fileSummary && fileSummary.bytes) {
```
max-zilla, line 51: loop through components here first

longshuicy (changes requested):
Like we discussed on Thursday, you might need to add similar logic in the useEffect.

max-zilla (approved):
Correctly shows when file is too big for bytes preview, when no viz config is available otherwise, and does not show either if a byte preview is shown. behaves as expected

longshuicy (approved):
Tested a few scenarios. Works and approve.

PR #725, opened 2023-09-12
fixing the problem where extractors always trigger

tcnichol:
This fixes the problem where extractors were triggering when they should not. The problem was that we were not checking the file type, just if the extractor was automatic.

### DIFF
--- a/backend/app/routers/feeds.py
+++ b/backend/app/routers/feeds.py
@@ -51,7 +51,6 @@ async def check_feed_listeners(
         feed_match = check_search_result(es_client, file_out, feed.search)
         if feed_match:
             for listener in feed.listeners:
-                # TODO is this necessary?
                 if listener.automatic:
                     listener_ids_found.append(listener.listener_id)
     for targ_listener in listener_ids_found:
--- a/backend/app/routers/feeds.py
+++ b/backend/app/routers/feeds.py
@@ -50,13 +50,10 @@ async def check_feed_listeners(
         # Verify whether resource_id is found when searching the specified criteria
         feed_match = check_search_result(es_client, file_out, feed.search)
         if feed_match:
-            feed_criteria = feed.search.criteria
-            for criterea in feed_criteria:
-                if criterea.field == 'content_type_main' and criterea.value == file_out.content_type.main_type:
-                    for listener in feed.listeners:
-                        # TODO not sure if this check is still needed
-                        if listener.automatic:
-                            listener_ids_found.append(listener.listener_id)
+            for listener in feed.listeners:
+                # TODO is this necessary?
+                if listener.automatic:
+                    listener_ids_found.append(listener.listener_id)
     for targ_listener in listener_ids_found:
         if (
             listener_info := await EventListenerDB.get(PydanticObjectId(targ_listener))
--- a/backend/app/routers/feeds.py
+++ b/backend/app/routers/feeds.py
@@ -50,9 +50,13 @@ async def check_feed_listeners(
         # Verify whether resource_id is found when searching the specified criteria
         feed_match = check_search_result(es_client, file_out, feed.search)
         if feed_match:
-            for listener in feed.listeners:
-                if listener.automatic:
-                    listener_ids_found.append(listener.listener_id)
+            feed_criteria = feed.search.criteria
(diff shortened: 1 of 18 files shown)


ddey2 (approved):
Works. looks good

backend/app/routers/feeds.py
```diff
@@ -50,9 +50,13 @@ async def check_feed_listeners(
         # Verify whether resource_id is found when searching the specified criteria
         feed_match = check_search_result(es_client, file_out, feed.search)
         if feed_match:
-            for listener in feed.listeners:
-                if listener.automatic:
-                    listener_ids_found.append(listener.listener_id)
+            feed_criteria = feed.search.criteria
+            for criterea in feed_criteria:
+                if criterea.field == 'content_type_main' and criterea.value == file_out.content_type.main_type:
```
max-zilla, line 55: this code is redundant to what is already being called on line 51, `check_search_result`. content_type and content_type_main are in the elasticsearch index and will be applied to the search for feed_match, if we are inside this block of code this should already be confirmed.

max-zilla (changes requested):
left comment about content type check

backend/app/routers/feeds.py
```diff
@@ -50,9 +50,13 @@ async def check_feed_listeners(
...
+                    for listener in feed.listeners:
+                        # TODO not sure if this check is still needed
+                        if listener.automatic:
```
longshuicy, line 58: Just curious, who's writing this "automatic" flag? Is it a part of the listener definition? If so could someone show an example?

longshuicy (changes requested):
Black linting is failing.

backend/app/routers/feeds.py
tcnichol: So for the query for JPEGS, here is the query string and here are the results:

`'{"preference":"results"}
{"query": {"bool": {"must": [{"match": {"name": "facebook_1633802270640_6852663398954784845.jpg"}}, {"bool": {"should": [{"match": {"content_type_main": "text"}}]}}]}}}

"{'took': 200, 'responses': [{'took': 200, 'timed_out': False, '_shards': {'total': 5, 'successful': 5, 'skipped': 0, 'failed': 0}, 'hits': {'total': {'value': 0, 'relation': 'eq'}, 'max_score': None, 'hits': []}, 'status': 200}]}"
'`

That is the correct query, but the conditional at the end always evaluated to True since it's checking the size of the entire response as a dictionary, and not whether the results had any hits. I have made another commit which should fix that.

I also sometimes got an error back from elasticsearch (timeout) and need to check if that comes up, or how this would be handled if it did.

backend/app/routers/feeds.py
max-zilla: it can be set when registering the extractor, or changed later. clowder v1 extractor_info will set this true by default.

backend/app/routers/feeds.py
```diff
@@ -51,6 +51,7 @@ async def check_feed_listeners(
         feed_match = check_search_result(es_client, file_out, feed.search)
         if feed_match:
             for listener in feed.listeners:
+                # TODO is this necessary?
```
max-zilla, line 54: yes, i think so. this will automatically trigger the listener/extractor on the newly uploaded file only if automatic is set to true, if it is false maybe the user only wants to send things to this extractor manually later.

max-zilla (approved):

max-zilla (changes requested):
changes make sense, need to fix pytest failing and lint formatting

tcnichol:
Looking at failing tests.

tcnichol:
A test is failing in test_feeds.

Looking at it, it's checking whether a file is being submitted to XYZ Test Feed. It is looking for a file with name xyz.

This is the query string, and it returns no results. 

```
'{"preference":"results"}
{"query": {"bool": {"must": [{"match": {"name": "xyz.txt"}}, {"bool": {"must": [{"match": {"name": "xyz"}}]}}]}}}
```

That looks like an impossible match (looking for a file with 2 exclusive conditions?) I am looking for a fix, but if anyone else has an idea, let me know.

max-zilla (approved):
tested with image previewer, triggers on images but not text or audio files.

PR #729, opened 2023-09-13
712 redesign the order of file upload form

longshuicy:
1. switch the order of metadata and file upload
2. fix the logic of checking required fields --> test by creating "required" and "not required" metadata, within each metadata you can further specify "required" field or "not required" field. The "next" button will enable only when "required" fields are all input
3. align the button to the right
Pasted Graphic 7
Pasted Graphic 6

### DIFF
--- a/frontend/src/components/files/UploadFile.tsx
+++ b/frontend/src/components/files/UploadFile.tsx
@@ -12,7 +12,7 @@ import {
 import { useDispatch, useSelector } from "react-redux";
 import { RootState } from "../../types/data";
 
-import { UploadFileModal } from "./UploadFileModal";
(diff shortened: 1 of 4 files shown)


lmarini:
@longshuicy looks good. Not directly related, but is the blue line under Choose File button required? If so, could the spacing between the button and the line be slightly increased?

longshuicy:
> @longshuicy looks good. Not directly related, but is the blue line under Choose File button required? If so, could the spacing between the button and the line be slightly increased?

I managed to disabled the underline. Pushed my commit.
image

max-zilla (approved):

PR #731, opened 2023-09-14
UI cleanup and bugfixes

max-zilla:
BUGFIXES
- UpdateFile now works on both the dataset file list and the individual file page. version auto-increments on successful upload but not on a failure. change versions endpoint to sort by latest so the limit of 20 wasn't hiding v21+.

CLEAN UP BUTTONS & SPACING (before/after)
dataset:
image

image

file:
image

image

groups:
image

image

Combine Users and Groups sharing lists on dataset into one table:
image


lmarini:
@max-zilla looks good. Some conflicts to resolve.

longshuicy (approved):
Looks good. If you can resolve those merge conflict we can merge

PR #736, opened 2023-09-17
735 sharing tab bugs

tcnichol:
This should fix both sharing tab bugs.

The error for both changing the user role and the group role were that the wrong arguments were getting passed to the remove_dataset_user_role and remove_dataset_group_role methods. That was causing indexing on elasticsearch to fail. Now both should be fixed.

### DIFF
--- a/backend/app/routers/authorization.py
+++ b/backend/app/routers/authorization.py
@@ -159,7 +159,7 @@ async def set_dataset_group_role(
                 )
             ) is not None:
                 if group_id not in auth_db.group_ids:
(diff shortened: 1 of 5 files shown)


tcnichol:
I added some new tests for sharing a dataset with a user and changing the role, and sharing the dataset with a group and changing the role.

They should catch similar bugs in the future.

longshuicy (approved):
Work as expected. Approve

ddey2 (approved):
works. Thanks for adding the tests

PR #739, opened 2023-09-20
702 thumbnail bytes sometimes doesnt show with 404

longshuicy:
Seems like that I need to store the cookie at the "root" directorate so files on the path can access it. 
Can't seem to trigger 404 anymore with this change; and I looked at the route that has that 404, nothing seems particularly wrong with it.


tcnichol:
Testing, I am not getting the 404 with any .png images, also tested with .jpg. 

I can mark this one approved.

tcnichol (approved):
Tested with png and jpg files, looks like the error is not happening anymore.

ddey2 (approved):

PR #743, opened 2023-09-24
701 improve file version selection

tcnichol:
Instead of clicking the version chip there is now an obvious and clear dropdown on the file page to make it clear that the user can select versions.

There is still a problem. The initial version is not selected, even though that is the version displayed. Will need to fix this.


longshuicy:
Sorry for nitpicking the aesthetics. This drop down might take too much space and might be too loud. Maybe we need to think an alternative placement? 
image

max-zilla:
Can we move the version pulldown to the details sidebar?

tcnichol:
> Can we move the version pulldown to the details sidebar?

I will move that and see how it looks.

tcnichol:
So I have tried to move the version select, but then it looks like all the FileDetails end up at the bottom, so the select isn't very visible:
Screenshot 2023-09-28 at 2 27 56 PM

longshuicy:
I cleaned up the file selection and added the snackbar
image
image

lmarini (approved):

PR #750, opened 2023-09-27
Enabling @context in metadata and metadata definition

ddey2:
This will enable to input metadata containing @context like jsonld and to show output with @context.This will also use @context while creating entry in mongodb.

To test
1. make sure test_metadata passes.
2. Create a metadata definition. Check in mogodb if it has created an entry with '@context'.
3. See in postman to get the metadata/ metadata defn you just created. It should show output containing '@context'

### DIFF
--- a/backend/app/models/metadata.py
+++ b/backend/app/models/metadata.py
@@ -101,6 +101,14 @@ class MetadataDefinitionBase(BaseModel):
     class Settings:
         name = "metadata_definitions"
 
+    class Config:
+        # Serialization Config Options
+        # Specify JSON key names
+        # This will rename the field `context` to `@context` in the JSON output
+        fields = {"context": {"alias": "@context"}}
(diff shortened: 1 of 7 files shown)


longshuicy (approved):
works well. I check the database, it's depositing it as @context.

backend/app/models/metadata.py
```diff
@@ -206,6 +222,12 @@ def definition_is_valid(cls, v):
     class Settings:
         name = "metadata"
 
+    class Config:
+        # Serialization Config Options
+        # Specify JSON key names
+        # This will rename the field `context` to `@context` in the JSON output
+        fields = {"context": "@context"}
```
longshuicy, line 229: Just curious, no need to actually say "alias"?

ddey2: Good catch Chen. This is an extra piece of code block. I had the Config class set up above. I removed this piece of code block in my next commit.

PR #751, opened 2023-09-27
746 on selecting date time picker widget type doesnt narrow down the field options

longshuicy:
1. Reconstruct the widget type and input type to tie them together
2. Style the form properly
3. Write more typescirpts

image


ddey2 (changes requested):
It works for the 1st case stated in the bug. I notice 1 more thing while tetsing this.

If you select Dropdown, then list, then selects Date&Time picker, 'supported list value' input box doesn't go away. 
Screenshot 2023-09-27 at 1 34 57 PM
Screenshot 2023-09-27 at 1 35 38 PM

tcnichol:
Tested and I can no longer pick integer or any other types. 

If I pick date I can still pick a time. Is that intended behavior?

longshuicy:
> Tested and I can no longer pick integer or any other types.
> 
> If I pick date I can still pick a time. Is that intended behavior?

Yes that's intended too restrict each widget type with its supported data type

tcnichol (approved):
Tested this one and marking approved.

PR #756, opened 2023-09-28
683 include thumbnail ids in elasticsearch index

tcnichol:
Thumbnails have changed to have a downloads field. This is not incremented by default at present. 

There is now an index_thumbnail method, though I am still using the file name since thumbnails do not have a name.

### DIFF
--- a/backend/app/routers/thumbnails.py
+++ b/backend/app/routers/thumbnails.py
@@ -73,7 +73,8 @@ async def remove_thumbnail(thumb_id: str, fs: Minio = Depends(dependencies.get_f
 
 @router.get("/{thumbnail_id}")
 async def download_thumbnail(
-    thumbnail_id: str, fs: Minio = Depends(dependencies.get_fs),
+    thumbnail_id: str,
+    fs: Minio = Depends(dependencies.get_fs),
     increment: Optional[bool] = False,
 ):
     # If thumbnail exists in MongoDB, download from Minio
--- a/backend/app/routers/thumbnails.py
+++ b/backend/app/routers/thumbnails.py
@@ -1,6 +1,7 @@
 from typing import List, Optional
 
 from beanie import PydanticObjectId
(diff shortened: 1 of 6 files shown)


tcnichol:
Making this draft, I need to check how this works with searches.

backend/app/routers/thumbnails.py
```diff
@@ -72,7 +73,9 @@ async def remove_thumbnail(thumb_id: str, fs: Minio = Depends(dependencies.get_f
 
 @router.get("/{thumbnail_id}")
 async def download_thumbnail(
-    thumbnail_id: str, fs: Minio = Depends(dependencies.get_fs)
+    thumbnail_id: str,
+    fs: Minio = Depends(dependencies.get_fs),
```
longshuicy, line 77: Is minio is being used?

longshuicy (commented):
Is the purpose of adding thumbnail to elasticsearch eventually leads to searchable thumbnail? 
Then would it possible to add this index to the search endpoint?

tcnichol:
> Is the purpose of adding thumbnail to elasticsearch eventually leads to searchable thumbnail? Then would it possible to add this index to the search endpoint?

From the issue Max created, I think the goal was that the thumbails would be returned with a seach. Like if you search for a file the thumbnail associated with that file would also be returned. I might need to ask Max.

backend/app/routers/thumbnails.py
tcnichol: Yes, it was used before. What I changed here is I added a downloads since to match the other fields in elasticsearch.

tcnichol: Also - the downloads are not incrementing because I was not counting them as downloading if they are just displayed. Wasn't sure if I did that right.

longshuicy (approved):

max-zilla (approved):

PR #762, opened 2023-10-02
Simple script to POST metadata definitions.

lmarini:
Also fixed syntax of one definition or it wouldn't POST.


ddey2 (approved):

longshuicy:
black is complaining about linting.. eventhough it's not in the backend folder it somehow still got checked.

PR #764, opened 2023-10-02
message on new metadata entry

tcnichol:
We want to show a success message when new metadata is created. 

I am starting with snackbar, and this is draft because I haven't got the colors fixed.


max-zilla (approved):

ddey2 (approved):
works

PR #767, opened 2023-10-03
message if no datasets exist, button link to create

tcnichol:
In an empty instance, we see a message telling us there are no datasets and a button that links to create datasets. This is a draft because I need to fix the positioning on the page.


ddey2 (approved):
looks good, tested and worked.

PR #786, opened 2023-10-10
Implement wordcloud visualization

ddey2:
Please checkout pyclowder branch in https://github.com/clowder-framework/pyclowder/pull/96

Steps to reproduce:

1. Upload an appropriate csv file (see attached csv file for example)
2. Run the extractor.
3. Look at visualization tab.
[frequent_bigrams.csv](https://github.com/clowder-framework/clowder2/files/12862682/frequent_bigrams.csv)
Screenshot 2023-10-10 at 5 55 02 PM

### DIFF
--- a/frontend/src/components/visualizations/VegaSpec/VegaSpec.tsx
+++ b/frontend/src/components/visualizations/VegaSpec/VegaSpec.tsx
@@ -23,7 +23,7 @@ const allowedType = [
 	"geojson",
 ];
 
-export default function WordCloud(props: TextProps) {
+export default function VegaSpec(props: TextProps) {
 	let { fileId, visualizationId } = props;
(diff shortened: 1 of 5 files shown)


tcnichol:
Question. I see the preview and it looks right in in the visualization tab. When I try to download the visualization or use a shared link, I get a JSON. That looks like intended behavior given the component, but just wanted to make sure.

If it is, I can mark this one approved.

ddey2:
@tcnichol that's because the vis data is a json (vega spec which is rendered using vega-lite). So, this is correct.

You can save the image as png. that should work.
Screenshot 2023-10-11 at 1 03 38 PM

tcnichol (approved):
Marked extractor approved. This visualization looks good and I can save as png and download.

PR #791, opened 2023-10-12
788 duplicated extractor registration when extractor version updated

tcnichol:
This is now fixed. If a new version is running, it replaces the existing version. 

The 'created' field is taken from the 'created' of the original extractor, since it seemed to make sense that 'created' should always be initial registration and 'modified' is the time of the latest version. 

In order to test, you will need to kill the docker container for the heartbeat_listener.py and run it locally.


longshuicy:
For some reason i still see two different versions

image

longshuicy (approved):
Tested and it works. 
For other tester, make sure you rebuilt the extractor-heartbeat image with this branch.

lmarini (approved):

PR #793, opened 2023-10-12
792 clear previous log before switching extraction logs

longshuicy:
When you switching between different extractors, those log should clear out immediately then reflect the new changes
image


ddey2 (approved):
code looks good. work well

PR #797, opened 2023-10-18
Pagination for files & folder page under dataset

ddey2:
Create a dataset with more than 10 files or folders. It should split them in pages containing 10files.


tcnichol:
I tested this with a script that uploaded 40 files. Looks good pagination works. Marking approved.

tcnichol (approved):
I wrote a script to upload 40 files, pagination worked back and forth, making approved.

longshuicy (approved):

PR #798, opened 2023-10-18
Updated the labels for Share

ddey2:
Screenshot 2023-10-18 at 12 50 37 PM


tcnichol:
I notice that "With" is capitalized for user but not group. I'm guessing these should be consistent but I'm not sure which words should be capitalized.

tcnichol (approved):
Text matches, marking approved.

longshuicy (approved):

PR #802, opened 2023-10-19
787 enable uploading multiple files at once

tcnichol:
This is a draft for now, but the idea is to allow for multiple files to be uploaded at once.

Currently you can select multiple files at the same time, but if you try to select them one after the other it just replaces the file. 

Also, I am getting a 422 error on upload. 

To see an example of the POST request that works, you can check the new upload_files in the test utils.


backend/app/routers/datasets.py
```diff
@@ -461,6 +461,48 @@ async def save_file(
...
+        if (dataset := await DatasetDB.get(PydanticObjectId(dataset_id))) is not None:
+            if user is None:
+                raise HTTPException(
+                    status_code=401, detail=f"User not found. Session might have expired."
+                )
+
+            new_file = FileDB(name=file.filename, creator=user, dataset_id=dataset.id)
+
+            if folder_id is not None:
+                if (folder := await FolderDB.get(PydanticObjectId(folder_id))) is not None:
```
max-zilla, line 486: you might consider moving this check (and the if dataset is not none check) outside the files loop so you only have to check dataset and create folder once instead of doing it repeatedly for each file

backend/app/tests/utils.py
```diff
@@ -160,6 +163,33 @@ def upload_file(
...
+    filenames=[filename_example_1, filename_example_2],
+    file_contents=[file_content_example_1, file_content_example_2]
+):
+    """Uploads a dummy file (optionally with custom name/content) to a dataset and returns the JSON."""
+    upload_files = []
+    uploads = []
+    for i in range(0, len(filenames)):
+        with open(filenames[i], "w") as tempf:
+            tempf.write(file_contents[i])
+            uploads.append(open(filenames[i], "rb"))
```
max-zilla, line 179: i think you just want filenames[i] in the list if you are calling os.remove() later

tcnichol:
Some recent changes.

In the DatasetService.ts, the formData.append() method never appended anything, so I am creating the formdata earlier. There are console logs that show that the formdata does have the right entries. 

However, when I actually get to making the request, I get a 400 error, and in chrome developer tools, under network, payload, the FormData appears to be blank. This is not the case with the single file.

backend/app/tests/utils.py
tcnichol: Should be fixed now.

ddey2 (approved):
Works. tested

longshuicy (approved):

PR #810, opened 2023-10-26
Added version to openapi docs and bottom

lmarini:
of frontend client. Removed Alpha release
from README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Clowder V2 (In Development - Alpha Release)
+# Clowder V2 (In Active Development)
 
 *For the previous version of Clowder, please see [Clowder V1](https://github.com/clowder-framework/clowder).*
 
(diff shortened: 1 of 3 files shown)


ddey2 (approved):
It looks good to me. 

One thing - did you plan to make the color change to grey? I think it's subtle enough but still there.

backend/app/routers/status.py
```diff
@@ -0,0 +1,12 @@
...
+from fastapi.security import HTTPBearer
+
+from app.models.status import Status
+
+router = APIRouter()
+security = HTTPBearer()
+
+
+@router.get("", response_model=Status)
+async def add_thumbnail():
```
longshuicy, line 11: sorry nitpicking. The function name has thumbnail. Do we want to change to status?
You might also need to run codegen after that

longshuicy (changes requested):
Just a minor comment otherwise it looks good.
Tested with basic create dataset, upload, download, add metadata etc, all works.

longshuicy (approved):

PR #815, opened 2023-10-31
804 move searchbox to top for metadata definition and group page

longshuicy:
I wrapped the search widget with title and prompt to a generic component.
Move it up to the top and clean up the groups and metadata definition page a little bit.

image
image


max-zilla (approved):

ddey2 (changes requested):
UI looks good. 

However the search is not working. I tried to search metadata definition. Lmk if that is not related to this PR. I ll create a separate bug.

ddey2 (approved):
I think it's bug. I have create a separate issue for that. Approving

PR #816, opened 2023-10-31
Register local files

max-zilla:
Adds a whitelist to the config and a couple separate functions/endpoints to register/delete/download a locally referenced file.

To test, add a local file path to the whitelist and POST a file from that path:

image

image


max-zilla:
Remaining frontend updates:
- hide Share button for local file storage
- hide Version if version = 0 (local file storage)
- add chip for storage type (local)

ddey2 (approved):
Works well. Tested. 

The code is easy to follow. Thanks for that!

tcnichol (changes requested):
Looks like it needs codegen run. Aside from that it looks fine, there is a pytest error but its keycloak so test just needs rerun.

max-zilla:
> Looks like it needs codegen run. Aside from that it looks fine, there is a pytest error but its keycloak so test just needs rerun.

should be fixed now.

backend/app/routers/files.py
```diff
@@ -155,17 +150,46 @@ async def add_file_entry(
...
+
+    content_type_obj = get_content_type(new_file.name, content_type)
+    new_file.content_type = content_type_obj
+    await new_file.insert()
+
+    # Add entry to the file index
+    await index_file(es, FileOut(**new_file.dict()))
+
+    # TODO - timing issue here, check_feed_listeners needs to happen asynchronously.
+    time.sleep(1)
```
longshuicy, line 176: just curious, isn't all those await already assume it's synchronous event, why do we need to still add 1 second of waiting time?

frontend/src/components/files/FileActionsMenu.tsx
```diff
@@ -74,8 +77,19 @@ export const FileActionsMenu = (props: FileActionsMenuProps): JSX.Element => {
 	const history = useNavigate();
 
 	const [confirmationOpen, setConfirmationOpen] = useState(false);
-
 	const [updateFileOpen, setUpdateFileOpen] = useState(false);
+	const [shareEnabled, setShareEnabled] = useState(false);
+
+	useEffect(() => {
+		if (storageType === "minio") {
+			console.log("share OK");
```
longshuicy, line 85: Sorry nitpicking, could you remove console log?

longshuicy (commented):
Upload and download both works.
Only thing I noticed is, when i switch to the "extracted metadata" tab I got Version 0 not exist error
image

backend/app/routers/files.py
tcnichol: I also get the file version error.
When files are uploaded a FileVersion object is added to the mongodb, but this does not happen for local files. Not sure what the best fix would be.

max-zilla: @longshuicy It seems there is a slight delay between indexing a file in ES and having ES properly return the file in a search. even though we await the insert, I was still having my search return 0 results if I didn't wait a second to perform the search. It's possible we could reduce this time or (better option) move the automatic triggering logic to some asynchronous process in the background unconnected to this endpoint.

max-zilla: Added a check to avoid looking up version 0 of metadata, for local etc. files that have version = 0 this is treated the same as None (get latest metadata instead). Should fix this for both getting/listing and posting/editing user metadata to lcoal files.

frontend/src/components/files/FileActionsMenu.tsx
max-zilla: done

tcnichol (approved):
I tested again since the fixes and this works. Marking approved.

backend/app/routers/datasets.py
```diff
@@ -509,6 +513,61 @@ async def save_files(
...
+            )
+
+        # Check to make sure file is cleared to proceed
+        cleared = False
+        for wpath in settings.LOCAL_WHITELIST:
+            if localfile_in.path.startswith(wpath):
+                cleared = True
+        if not cleared:
+            raise HTTPException(
+                status_code=401,
```
lmarini, line 539: Should this be a 500 since it's an internal error? Or something less generic? 401 confused be as unauthorized.

max-zilla: changed this to 500

backend/app/config.py
```diff
@@ -35,7 +35,15 @@ class Settings(BaseSettings):
     MINIO_EXPIRES: int = 3600  # seconds
     MINIO_SECURE: str = "False"  # http vs https
 
-    # keycloak server
+    # Local file storage information
+    # Uploaded files will be written to a local folder instead of Minio.
+    LOCAL_STORAGE_ENABLED: bool = False
+    LOCAL_STORAGE_DIR: str = "/data"
```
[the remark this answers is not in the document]
max-zilla: removed

backend/app/routers/files.py
```diff
@@ -155,17 +150,46 @@ async def add_file_entry(
...
 
 
+async def add_local_file_entry(
+    new_file: FileDB,
+    user: UserOut,
+    es: Elasticsearch,
+    rabbitmq_client: BlockingChannel,
+    content_type: Optional[str] = None,
+):
+    """Insert FileDB object into MongoDB (makes Clowder ID), then Minio (makes version ID), then update MongoDB with
```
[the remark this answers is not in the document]
max-zilla: removed, thanks

backend/app/config.py
```diff
@@ -35,7 +35,15 @@ class Settings(BaseSettings):
     MINIO_EXPIRES: int = 3600  # seconds
     MINIO_SECURE: str = "False"  # http vs https
 
-    # keycloak server
+    # Local file storage information
+    # Uploaded files will be written to a local folder instead of Minio.
+    LOCAL_STORAGE_ENABLED: bool = False
```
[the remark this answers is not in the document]
max-zilla: removed

backend/app/routers/files.py
```diff
@@ -155,17 +150,46 @@ async def add_file_entry(
     )
 
 
+async def add_local_file_entry(
```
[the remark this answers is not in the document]
max-zilla: yes, there were comments to that effect on a couple of the functions. made an issue, maybe I can do this as part of the remote URL piece: https://github.com/clowder-framework/clowder2/issues/849

PR #832, opened 2023-11-15
485 public datasets visible when not logged in

tcnichol:
ready for review now


lmarini (changes requested):
Haven't looked at the code yet, but wanted to write down a few comments on the functionality:

- Error downloading dataset from public page (works in private)
- File page doesn't show details component on right hand side
- Extraction history tab looks different from when logged in. Probably permission. Do we want to even show this? I am not sure a public users needs this info. Remove tab?

tcnichol:
> Haven't looked at the code yet, but wanted to write down a few comments on the functionality:
> 
> * Error downloading dataset from public page (works in private)
> * File page doesn't show details component on right hand side
> * Extraction history tab looks different from when logged in. Probably permission. Do we want to even show this? I am not sure a public users needs this info. Remove tab?

I have removed extraction history from the public views. The public file page should also show the details properly now.

With download, still looking for a fix. Noticed something strange - repeatedly clicking the download button will eventually download a public dataset while not logged in, but one click has no effect. Working on this.

tcnichol:
@lmarini i didn't do anything specifically to fix download, but it now appears to work on one click. The dataset I made public locally took a while, but I also realize I put some rather large files in it. If you can check again (and the other issues) let me know.

backend/app/routers/authorization.py
```diff
@@ -1,3 +1,5 @@
+import datetime
```
longshuicy, line 1: any reason you need date time?

backend/app/models/files.py
```diff
@@ -95,6 +109,7 @@ class FileDBViewList(View, FileBase):
     bytes: int = 0
     content_type: ContentType = ContentType()
     thumbnail_id: Optional[PydanticObjectId] = None
+    status: str = FileStatus.PRIVATE.name
```
longshuicy, line 113: you already have it in the "FileBase", maybe no need to repeat here unless it's necessary?

backend/app/routers/datasets.py
```diff
@@ -522,6 +530,12 @@ async def save_files(
                         status_code=404, detail=f"Folder {folder_id} not found"
                     )
 
+            public = False
+            authenticated = False
+            if dataset.status == "PUBLIC":
+                public = True
+            if dataset.status == "AUTHENTICATED":
```
longshuicy, line 537: Could you help me understand the concept of "AUTHENTICATED"?

backend/app/main.py
```diff
@@ -164,6 +185,11 @@
     tags=["elasticsearch"],
     dependencies=[Depends(get_current_username)],
 )
+api_router.include_router(
+    elasticsearch.router,
```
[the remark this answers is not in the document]
longshuicy: I see you have `public_elasticsearch.py` file, maybe it's a typo?

backend/app/main.py
```diff
@@ -182,6 +208,11 @@
     tags=["visualizations"],
     dependencies=[Depends(get_current_username)],
 )
+api_router.include_router(
+    visualization.router,
```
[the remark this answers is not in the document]
longshuicy: same did you mean `public_visualization`?

backend/app/routers/public_datasets.py
```diff
@@ -0,0 +1,353 @@
+import datetime
```
longshuicy, line 1: Just curious is there any similarity to `datasets.py`?

frontend/src/routes.tsx
```diff
@@ -98,6 +101,12 @@ export const AppRoutes = (): JSX.Element => {
 	return (
 		<BrowserRouter>
 			<Routes>
+				<Route
```
longshuicy, line 104: did we settle on using a separate route of /public?

frontend/src/utils/common.js
```diff
@@ -52,6 +52,25 @@ export async function downloadResource(url) {
 	}
 }
 
+export async function downloadPublicResource(url) {
```
longshuicy, line 55: could we just reuse the downloadResource function? the content is exactly the same except leave out the "authHeader"

longshuicy (changes requested):

backend/app/routers/datasets.py
tcnichol: If a dataset is "public" then anyone can see it, even if they are not logged in.

If a dataset is "authenticated" then anyone with an account who is logged in can see it.

.run/uvicorn.run.xml
```diff
@@ -7,8 +7,9 @@
       <env name="PYTHONUNBUFFERED" value="1" />
     </envs>
     <option name="SDK_HOME" value="" />
+    <option name="SDK_NAME" value="Pipenv (clowder2)" />
     <option name="WORKING_DIRECTORY" value="$PROJECT_DIR$/backend" />
-    <option name="IS_MODULE_SDK" value="true" />
+    <option name="IS_MODULE_SDK" value="false" />
```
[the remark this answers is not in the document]
tcnichol: reverted.

frontend/src/routes.tsx
tcnichol: Yes, this one does not have the Depends(get_current_user) for methods but will always only return datasets that are public, and does not have routes that edit or create datasets. 

I did think that it might have been a good idea to have made a dependency that checks public, but it didn't seem absolutely necesary.

backend/app/models/files.py
tcnichol: I removed this, it was not needed.

frontend/src/utils/common.js
tcnichol: I tested and I could use downloadResource, removed the other

backend/app/routers/datasets.py
longshuicy: thanks for the explanation!

backend/app/deps/authorization_deps.py
```diff
@@ -227,7 +254,15 @@ async def __call__(
                     detail=f"User `{current_user} does not have `{self.role}` permission on file {file_id}",
                 )
             else:
-                raise HTTPException(status_code=404, detail=f"File {file_id} not found")
+                public_access = await check_public_access(
+                    file_id, "file", self.role, current_user
+                )
+                if public_access:
+                    return True
```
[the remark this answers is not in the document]
tcnichol: I did it like this since it looked like the method was supposed to either return True or throw an error. But I did change it to returning public_access instead of 'True' since that looked confusing.

backend/app/routers/public_datasets.py
tcnichol: it has some of the methods of datasets but not the Depends(get_current_user) and it does not have a dependency added to the router in main.

It only has the get methods.

backend/app/deps/authorization_deps.py
```diff
@@ -6,13 +6,40 @@
...
 from app.models.metadata import MetadataDB
 from app.models.pyobjectid import PyObjectId
 from app.routers.authentication import get_admin
 
 
+async def check_public_access(
+    resource_id: str,
+    resource_type: str,
+    role: RoleType,
+    current_user=Depends(get_current_username),
```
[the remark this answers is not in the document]
tcnichol: Let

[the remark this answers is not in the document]
tcnichol: Let me do some testing here.

backend/app/deps/authorization_deps.py
```diff
@@ -6,13 +6,40 @@
...
+) -> bool:
+    has_public_access = False
+    if role == RoleType.VIEWER:
```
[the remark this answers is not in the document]
tcnichol: The method is only called in cases where the user did not have a specific role on a dataset.

If the role being requested was not Viewer, check_public_access should return False, even if the dataset status is 'public.' 

Passing in the role might be confusing, so I might try to think of a better way to handle this. I was just going for a method that would be easy to reuse.

frontend/src/routes.tsx
tcnichol: I just pushed a change that I thought would work for this, but so far it still redirects to keycloak. Working on this.

tcnichol: I think this is fixed now.

tcnichol:
Some newer changes.

In frontend/components/visualizations there is logic that checks if we are looking at a public component, and then uses a different route to download the file and get visualizations depending. The visualizations were not showing up if a user was not logged in.

backend/app/deps/authorization_deps.py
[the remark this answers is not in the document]
tcnichol: I'm going to leave this for now, I cannot think of a better way.

backend/app/deps/authorization_deps.py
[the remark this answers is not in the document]
tcnichol: The reason we have the user is that if the user is logged in, they should still see PUBLIC or AUTHENTICATED datasets. That's why it was done like this.

backend/app/deps/authorization_deps.py
```diff
@@ -41,24 +73,11 @@ async def get_role_by_file(
...
-                    raise HTTPException(
-                        status_code=403,
-                        detail=f"User `{current_user} does not have role on file {file_id}",
-                    )
+        public_access = await check_public_access(
+            file_id, "file", RoleType.VIEWER, current_user
+        )
+        if authorization is None and public_access:
+            return RoleType.VIEWER
         return authorization.role
```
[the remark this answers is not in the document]
tcnichol: Yes, that is changed. It looks like it's supposed to either return a role or throw an error, not return None

longshuicy:
image

The "Search" on the left drawer doesn't work. I think search is a protected endpoint. Do we want to remove that altogether?

longshuicy:
Same as the search bar on top
image

backend/app/deps/authorization_deps.py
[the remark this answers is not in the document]
longshuicy: There is still some quirks here. For file You might still want to catch the case that "authorization" is None but public_access is "false"

I got an error that traced to this place. Try: 
- mark an empty dataset "public"
- upload a file to that dataset.
- go to that file
- I get an error 
```
 File "/Users/cwang138/.local/share/virtualenvs/backend-pkZ6g-7g/lib/python3.9/site-packages/fastapi/dependencies/utils.py", line 622, in solve_dependencies
    solved = await call(**sub_values)
  File "/Users/cwang138/Documents/Projects/Clowder/clowder2/backend/app/deps/authorization_deps.py", line 81, in get_role_by_file
    return authorization.role
AttributeError: 'NoneType' object has no attribute 'role'
```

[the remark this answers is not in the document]
longshuicy: image

longshuicy (changes requested):
image
 Metadata is protected as well

frontend/src/actions/public_dataset.js
```diff
@@ -0,0 +1,144 @@
...
+			.catch((reason) => {
+				dispatch(handleErrors(reason, fetchPublicDatasetAbout(id)));
+			});
+	};
+}
+
+export const RECEIVE_PUBLIC_DATASETS = "RECEIVE_PUBLIC_DATASETS";
+
+export function fetchPublicDatasets(skip = 0, limit = 21) {
+	console.log("we are here");
```
longshuicy, line 101: Could you remove the console.log?

frontend/src/components/visualizations/Image/Image.tsx
```diff
@@ -2,25 +2,37 @@ import React, { useEffect, useState } from "react";
...
+			}
 		} else {
-			downloadUrl = generateFileDownloadUrl(fileId, 0);
+			if (publicView){
+				downloadUrl = generatePublicFileDownloadUrl(fileId, 0);
+			} else {
+				downloadUrl = generateFileDownloadUrl(fileId, 0);
+			}
 		}
+		console.log('download URL is', downloadUrl);
```
longshuicy, line 35: Could you remove the console.log?

frontend/src/components/visualizations/PublicVisualization.tsx
```diff
@@ -0,0 +1,184 @@
...
+		if (fileId !== undefined) {
+			listPublicFileSummary(fileId);
+			getPublicVisConfig(fileId);
+		}
+
+		if (datasetId !== undefined) {
+			getPublicVisConfig(datasetId);
+		}
+	}, [fileId, datasetId]);
+	console.log("file summary is", fileSummary);
```
longshuicy, line 51: Could you remove the console.log?

frontend/src/components/visualizations/PublicVisualizationCard.tsx
```diff
@@ -0,0 +1,90 @@
...
+type publicPreviewProps = {
+	publicVisComponentDefinition: VisComponentDefinitions;
+	publicVisualizationDataItem: VisualizationDataOut;
+};
+
+export const PublicVisualizationCard = (props: publicPreviewProps) => {
+	const { publicVisComponentDefinition, publicVisualizationDataItem } = props;
+	const [expanded, setExpanded] = React.useState(false);
+	const [visShareModalOpen, setVisShareModalOpen] = useState(false);
+	console.log("public visualization card");
```
longshuicy, line 25: Could you remove the console.log?

frontend/src/utils/common.js
```diff
@@ -31,6 +31,26 @@ export function getHeader() {
...
+	const authHeader = getHeader();
+	console.log('downloading public resource', url);
```
longshuicy, line 36: Could you remove the console.log?

frontend/src/components/files/FilesTableFileEntry.tsx
```diff
@@ -16,13 +16,14 @@ type FilesTableFileEntryProps = {
...
+	publicView: boolean | false;
 };
 
 export function FilesTableFileEntry(props: FilesTableFileEntryProps) {
-	const { iconStyle, selectFile, file, parentFolderId } = props;
+	const { iconStyle, selectFile, file, parentFolderId, publicView } = props;
 	const [thumbnailUrl, setThumbnailUrl] = useState("");
 	const [selectedVersion, setSelectedVersion] = useState(file.version_num);
-
+	console.log('the public view is', publicView);
```
longshuicy, line 26: Could you remove the console.log?

longshuicy:
After reading your code, I realize you probably need to point those frontend link/buttons to the "public_xxx" endpoint, e.g. search, metadata, etc. Could you double check if there is anything else missing?

frontend/src/components/visualizations/Image/Image.tsx
tcnichol: removed

frontend/src/actions/public_dataset.js
tcnichol: removed

frontend/src/components/files/FilesTableFileEntry.tsx
tcnichol: removed

frontend/src/utils/common.js
tcnichol: removed

frontend/src/components/visualizations/PublicVisualizationCard.tsx
tcnichol: removed

frontend/src/components/visualizations/PublicVisualization.tsx
tcnichol: removed

backend/app/deps/authorization_deps.py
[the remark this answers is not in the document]
tcnichol: @longshuicy I could not replicate this, but I did add throwing an exception. Let me know if this works.

ddey2:
Few things I noticed:

1. Search is working now. 
2. I can upload file to a public dataset. 
3. If I try to view 'User metadata' tab, I get error. 
Screenshot 2024-01-16 at 2 36 28 PM
4. If I got to 'sharing tab', I get 'Forbidden page' error. Sharing doesn't work.

ddey2 (changes requested):
Left the things I noted in comment

lmarini:
@tcnichol can you update the description of the PR to describe what it does? Not just for the current reviewers but for future reference as well. Thank you.

tcnichol:
How to test this pull request.

Note - the new 'default' page is now the public page. Login and register are in the upper right hand corner.

Create at least 2 Users. 

Using User1, create datasets and upload files. Then mark at least one dataset 'public' by changing the status from 'private' to 'public.' 

If you log out as that user, then you will see the public datasets. Other users will also see them when logged in. So if User1 made a dataset public, User2 will see it when logged in.

tcnichol:
New instructions for testing.

If you start up a clowder instance, run the populate fake data script. It will generate some public datasets and they should be visible when you are not logged in.

The new default page is the public page, with login and register in the upper right corner.

frontend/src/actions/public_file.js
```diff
@@ -0,0 +1,209 @@
...
+			.catch((reason) => {
+				dispatch(handleErrors(reason, fetchPublicFileSummary(id)));
+			});
+	};
+}
+
+export const RECEIVE_PUBLIC_FILE_METADATA_JSONLD = "RECEIVE_PUBLIC_FILE_METADATA_JSONLD";
+
+export function fetchPublicFileMetadataJsonld(id) {
+	const url = `${config.hostname}/public/files/${id}/metadata.jsonld`;
```
longshuicy, line 74: Hmm do we even have a jsonld endpoint? you probably can leave this out

frontend/src/actions/public_file.js
```diff
@@ -0,0 +1,209 @@
...
+				dispatch(handleErrors(reason, fetchPublicFileMetadataJsonld(id)));
+			});
+	};
+}
+
+export const RECEIVE_PUBLIC_PREVIEWS = "RECEIVE_PUBLIC_PREVIEWS";
+
+export function fetchPublicFilePreviews(id) {
+	const url = `${config.hostname}/public/files/${id}/getPreviews`;
+	return (dispatch) => {
```
longshuicy, line 100: same here we don't have endpoint for that. That's most likely just for v1 clowder

longshuicy (changes requested):
### Could we remove those dividers?
image
---
### Download version history doesn't work. i think the url is wrong, instead of public/files you should use public_files in the URL
image
image
---
### Download dataset button doesn't work. It's still pointing to the private endpoint.
image
image
---
### Download dataset button inside the dataset page doesn't work either. Not sure why
image
image

frontend/src/actions/public_file.js
```diff
@@ -0,0 +1,209 @@
...
+			});
+	};
+}
+
+export const RECEIVE_PUBLIC_FILE_EXTRACTED_METADATA =
+	"RECEIVE_PUBLIC_FILE_EXTRACTED_METADATA";
+
+export function fetchPublicFileExtractedMetadata(id) {
+	const url = `${config.hostname}/api/v2/public/files/${id}/metadata`;
+	return (dispatch) => {
```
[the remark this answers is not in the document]
tcnichol: I tried to replicate what was done in actions/file.js with this one. Maybe both should be changed?

frontend/src/actions/public_file.js
tcnichol: removed.

frontend/src/actions/public_file.js
tcnichol: removed.

tcnichol:
@longshuicy i removed the divider and addressed your comments.
There is one method that i left a comment on - in public_file I modeled the metadata function that gets extracted metadata after the one in actions/file.js. Maybe both should change?

frontend/src/actions/public_file.js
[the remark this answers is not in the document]
longshuicy: I dont think this is being used anywhere. Maybe just legacy code
image

longshuicy (approved):
All the suggestions are resolved now. Approve. 
But could you fix the black linting before merging?

tcnichol:
I'm not sure what is going wrong.

I am running

`pipenv run black develop`

and i see that it's changing populate_fake_data.py

but when I go to commit it shows nothing. 

Same with the backend/app

PR #834, opened 2023-11-17
830 create superadmin mode

ddey2:
1. show if an user is admin or not on UI
2. Enable admin user to view everything if only the admin_mode is turned on


tcnichol:
I pushed a small fix in authorizationservice.

Testing it, I can see all users datasets and files with admin mode activated, and when turned off we revert to what it was before.

tcnichol (approved):
With the last few commits all the issues are resolved and this works. Marking approved.

PR #841, opened 2023-11-30
824 display more information on the extractor popup

longshuicy:
image
image
image

### DIFF
--- a/frontend/src/components/listeners/ExtractionHistory.tsx
+++ b/frontend/src/components/listeners/ExtractionHistory.tsx
@@ -11,7 +11,8 @@ import {
 	ListSubheader,
 } from "@mui/material";
 import { useDispatch, useSelector } from "react-redux";
-import { Listener, RootState } from "../../types/data";
+import { RootState } from "../../types/data";
+import { EventListenerOut as Listener } from "../../openapi/v2";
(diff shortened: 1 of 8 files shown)


lmarini (changes requested):
Looks great. Three minor suggestions:
- Could we get a close button or x in top right for popup?
- The tables on the files and the extraction history tabs have borders. To be consistent should we have one here as well?
- Not changed in this PR, but when an extractor is not alive, should we still be able to click on the title and get the description? Right now we can't.
Thanks!

max-zilla (approved):

longshuicy:
> Looks great. Three minor suggestions:
> 
> * Could we get a close button or x in top right for popup?
> * The tables on the files and the extraction history tabs have borders. To be consistent should we have one here as well?
> * Not changed in this PR, but when an extractor is not alive, should we still be able to click on the title and get the description? Right now we can't.
>   Thanks!

I think i addressed all of them. i also rewrote the searchbox (using generic search component we have), as well as align them better. 
image
image

tcnichol (approved):
This looks good. I like the expanded info on both the extractor page and on submit. 

Marking approved.

lmarini (approved):

PR #845, opened 2023-12-04
842 UI component for assign admin

longshuicy:
image

-----

### This PR:
1. Add endpoint to revoke admin privilege
2. Logic to forbid current "admin" user to revoke their own admin access
3. Admin field was add to the wrong model, fixed that.
5. Frontend page to manage users, search users, and set/revoke admin access

### To test:
1. Run the `scripts/develop/populate_fake_data/populate_fake_data.py`. Change number of users if needed
2. Admin user will have "Manage Users" option at the top bar menu. (Normal user won't)
3. Try set/revoke admins and watch the page 

Pasted Graphic 1
Pasted Graphic


lmarini (changes requested):
Works great. I still had to drop datasets and metadata definition to make it work with the admin field. I also added a line to delete-data.sh to drop metadata definition.

Should we move the "Manage Data" link from the personal menu in the top right to the tray menu on the left?

longshuicy:
> Works great. I still had to drop datasets and metadata definition to make it work with the admin field. I also added a line to delete-data.sh to drop metadata definition.
> 
> Should we move the "Manage Data" link from the personal menu in the top right to the tray menu on the left?

Ah i see that's what's missing in the drop script.

Do mean the "Manage Users" link? I can certainly do that :-)

longshuicy:
> Works great. I still had to drop datasets and metadata definition to make it work with the admin field. I also added a line to delete-data.sh to drop metadata definition.
> 
> Should we move the "Manage Data" link from the personal menu in the top right to the tray menu on the left?

Moved to the side drawer now 
image

tcnichol:
Tested this one and everything looks good. Users cannot revoke their own status.

tcnichol (approved):
Tested this one and everything seems to work fine. Checked that users could not remove own admin status.

PR #846, opened 2023-12-04
Pre commit autoformat

lmarini:
This PR introduces pre-commit hooks to run:

- [x] some python formatting checks
- [x] generation of openapi spec 
- [x] codegen of javascript client stubs
- [x] javascript eslint and prettier

To setup install python library with `pipenv install` and setup hooks with `pre-commit install`. 

*NOTE:* once in place, hooks are in the `.git` directory and will be run on any branch. They can be removed with `pre-commit uninstall` or by skipping them when git commit with `--no-verify`. 

Use `pre-commit run --all-files` to run them before committing.


.pre-commit-config.yaml
```diff
@@ -0,0 +1,53 @@
...
+      - id: extract-openapi
+        name: extract-openapi
+        entry: pipenv run extract-openapi
+        language: system
+        pass_filenames: false
+  - repo: local
+    hooks:
+      - id: generate-frontend-api-client
+        name: generate-frontend-api-client
+        entry: bash -c 'cd frontend && npm run codegen:v2:file'
```
longshuicy, line 13: What's the trick that makes this command work without backend running?

longshuicy: oh im answering myself...
Is it because the script `extract-openapi = "python backend/extract-openapi.py app.main:app"` can generate openapi.json without actually having a running background?
Does that mean I can also update the github action to eliminate all the steps of starting up all docker stacks?

longshuicy (approved):
Followed the steps and everything works. Approve.

.pre-commit-config.yaml
lmarini: correct

PR #858, opened 2023-12-19
allow message and heartbeat image and tag to be overwrite

longshuicy:


deployments/kubernetes/charts/clowder2/values.yaml
```diff
@@ -168,25 +168,31 @@ rabbitmq:
     #storageClass: csi-cinder-sc-retain
 
 # ----------------------------------------------------------------------
-# EXTRACTORS
+# DEFAULT EXTRACTOR SERVICES
 # ----------------------------------------------------------------------
+message:
+  image:
+    repository: clowder/clowder2-messages
+    tag: main
```
robkooper, line 176: better to comment these out, otherwise it will not default to the chart.appversion in your templates by defaul. Best to have people set this in the overrides.

robkooper, line 176: better to comment these out, otherwise it will not default to the chart.appversion in your templates by defaul. Best to have people set this in the overrides.

deployments/kubernetes/charts/clowder2/values.yaml
```diff
@@ -168,25 +168,31 @@ rabbitmq:
...
+# EXTRACTORS
+# ----------------------------------------------------------------------
 ## list of extractors to be installed. Each extractor should have the
 ## following fields:
 ## enabled      : true/false - should the extractor be installed
 ## image        : dokeer/image:tag - the extractor docker image
 ## replicaCount : number - the number of this extractor to install
 ## env          : additional environment variables to pass to extractor.
 extractors:
-#  file-digest:
```
robkooper, line 181: any specific reason to remove digest?

robkooper (changes requested):
better to comment out the `tag: main`

deployments/kubernetes/charts/clowder2/values.yaml
longshuicy: we haven't converted to be v2 compatible yet.. maybe I should create a task for that.

lmarini (approved):

PR #859, opened 2024-01-03
Configure memory limit on ElasticSearch container in development

blinkdog:
Adds a small stanza to `docker-compose.dev.yml` to limit ElasticSearch container to 2 GB of memory.
Add `--compatibility` flag to `docker-dev.sh` to allow Docker Compose to make use of the configured limit.

Reference: https://stackoverflow.com/a/57135933


blinkdog:
Sorry for the noise.

PR #860, opened 2024-01-03
adding timeout loop for heartbeat and messages

tcnichol:
This is a simple loop.

I saw some interesting timeout features in libraries like Threading, but that seemed more complicated than we needed for this.


ddey2 (approved):
logic is simple and straight forward.

lmarini (approved):

PR #862, opened 2024-01-08
861 bug when preferred username is not the email

tcnichol:
ready for review. look at @longshuicy comment below to add the preferred_username

### DIFF
--- a/backend/app/keycloak_auth.py
+++ b/backend/app/keycloak_auth.py
@@ -329,6 +329,7 @@ async def create_user(email: str, password: str, firstName: str, lastName: str):
         {
             "email": email,
             "username": email,
+            "preferred_username": email,
             "enabled": True,
             "firstName": firstName,
             "lastName": lastName,
--- a/backend/app/keycloak_auth.py
+++ b/backend/app/keycloak_auth.py
@@ -329,7 +329,6 @@ async def create_user(email: str, password: str, firstName: str, lastName: str):
         {
             "email": email,
             "username": email,
-            "preferred_username": email,
             "enabled": True,
             "firstName": firstName,
             "lastName": lastName,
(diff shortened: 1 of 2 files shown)


tcnichol:
To replicate this bug, follow the instructions from this link to add a preferred_username field to a user.

https://medium.com/@ramanamuttana/custom-attribute-in-keycloak-access-token-831b4be7384a

Once you do that, and you try to log in as that user, you will see 'error user {preferred_username} not found.' on main. On this branch it should be fixed.

lmarini:
> To replicate this bug, follow the instructions from this link to add a preferred_username field to a user.
> 
> https://medium.com/@ramanamuttana/custom-attribute-in-keycloak-access-token-831b4be7384a
> 
> Once you do that, and you try to log in as that user, you will see 'error user {preferred_username} not found.' on main. On this branch it should be fixed.

PR looks good, but I am not sure I understand this comment. I think preferred_username is always there

tcnichol:
@lmarini 

When I created users they didn't have the preferred_username field, so I found a way to add it so I could have the preferred_username not be the email and replicate the bug and test the fix. If you don't need to, then just edit the preferred_username in keycloak to make sure it's not the same as email. On main you'll see an error, but on this branch you won't.

longshuicy:
How can I test this PR?

tcnichol:
To test the PR, follow the instructions in the linked article to add "preferred_username" field to a user. Set the preferred username to something aside from the email. For me, users did not have 'preferred_username' field if they were created through clowder. 

On main, you'll see an error when logging in with that user, but on this branch the error is fixed.

You can also just add the field for 1 user.

longshuicy (approved):
Code looks good and works.

I think i know how to reproduce the error:
1. Go to clients --> find clowder2-backend --> select "client scopes"
2. select clowder2-backend-dedicate
3. Mapper --> add mapper by configuration
4. create a user attribute named "preferred_username"
5. go back to a given user, add the "preferred_username" as a attribute and give it a different name
6. You should expect failure with "main" branch; but success with this PR fix

image
image
image
image

alemsh:
Also confirming that this change work with our keycloak environment as well, thanks!

PR #863, opened 2024-01-08
hide files pagination on other tabs

tcnichol:
I put this inside a conditional. We just see prev and next for files and then for extractors.

### DIFF
--- a/backend/app/deps/authorization_deps.py
+++ b/backend/app/deps/authorization_deps.py
@@ -1,24 +1,29 @@
 from beanie import PydanticObjectId
 from beanie.operators import Or
-from bson import ObjectId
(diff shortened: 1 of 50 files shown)


lmarini (changes requested):
I don't think we need a conditional? If it's just supposed to be in the first TabPanel, can we just move it there? This was probably just a bad merge.

tcnichol:
> I don't think we need a conditional? If it's just supposed to be in the first TabPanel, can we just move it there? This was probably just a bad merge.

I made the change.

lmarini (approved):

longshuicy (approved):
Looks good.

PR #888, opened 2024-01-18
add metadata button is back

tcnichol:
fixes the missing metadata button.


ddey2 (approved):
looks good

tcnichol:
The metadata button was missing since setEnableAddMetadata should not have been added to the conditional for the button. It starts false and will be set to true by the button.

frontend/src/components/files/File.tsx
```diff
@@ -275,7 +275,7 @@ export const File = (): JSX.Element => {
 	} else if (showNotFoundPage) {
 		return <PageNotFound />;
 	}
-
+	console.log("file role is", fileRole);
```
longshuicy, line 278: Could you remove that logg?

longshuicy:
After I created some user metadata and then click "add metadata" it leads me to a blank page
image
image

tcnichol:
@longshuicy 

I have not been able to replicate the blank screen after add metadata.

I tried to replicate the same situation - metadata type abstract and metadata text field, but it worked for me and the new metadata was added to the dataset.

tcnichol:
I am able to replicate the blank screen error. Working to fix.

tcnichol:
I think I fixed the empty screen problem after clicking add metadata.

datasetRole !== undefined works
datasetRole.role !== undefined does not, since you can't get .role from undefined.

frontend/src/components/metadata/widgets/MetadataTextField.tsx
```diff
@@ -39,7 +39,7 @@ export const MetadataTextField = (props) => {
 				/>
 			</Grid>
 			<Grid item xs={1} sm={1} md={1} lg={1} xl={1}>
-				{datasetRole.role !== undefined && datasetRole.role !== "viewer" ?
+				{datasetRole !== undefined && datasetRole.role !== "viewer" ?
```
longshuicy, line 42: you might want to keep the original logic?
`datasetRole !== undefined && datasetRole.role !== undefined && datasetRole.role !== "viewer`

ddey2: We definitely need the whole check as Chen mentioned. else it still shows blank page.

ddey2 (changes requested):

frontend/src/components/metadata/widgets/MetadataTextField.tsx
tcnichol: It is back in that order.

longshuicy (approved):

PR #897, opened 2024-01-23
upgrade keycloak to v20

tcnichol:
With keycloak v20, openid is not included as a default scope, which caused the error reported by the upgrade.

On the issue there is a video that shows how I configured keycloak through the admi console to work.

I did an export of the realm.json and I believe that all the keycloak json files are now working and correct. The dev one definitely works. 

Marking this draft as I need to check whether the other jsons work.

### DIFF
--- a/deployments/kubernetes/charts/clowder2/files/realm.json
+++ b/deployments/kubernetes/charts/clowder2/files/realm.json
@@ -197,7 +197,8 @@
         "acr",
         "roles",
         "profile",
-        "email"
+        "email",
+        "openid"
       ],
       "optionalClientScopes": [
         "address",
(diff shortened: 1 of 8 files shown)


tcnichol:
I have tested this with both docker-dev and docker compose and it works.

I was not sure what to do to test other keycloak json files that I modified, but I will make this a pull request and not draft.

longshuicy (commented):
Could you also update the k8s deployment realm at deployments/kubernetes/charts/clowder2/files/realm.json if possible?

tcnichol:
> Could you also update the k8s deployment realm at deployments/kubernetes/charts/clowder2/files/realm.json if possible?

Added.

longshuicy (approved):

Issue #901, opened 2024-01-29
Pagination update with public datasets


longshuicy:
Backend and frontend needs to update to the new pagination

PR #903, opened 2024-01-31
896 folder plus file pagination

longshuicy:
This PR features:
1. Create additional view that combines both file and folders
2. Add indicator "object_type" to differentiate 
3. Additional endpoint `/dataset/{id}/folders_and_files` to list all the folders and files; with folders listed first and sorted by created date
4. Frontend listing, pagination logic
5. A few clean ups

-----
To Test:
1. Try create nested folder structures
2. Upload files within each level see if it works
3. Try delete folder, delete files, update files and etc to see everything works

-----
To Do:
I haven't touch anything related to public datasets. Will address in another issue: #901


tcnichol (approved):
I tested this for non-public datasets. It looks good and works, so I am marking approved.

PR #912, opened 2024-02-05
set up helm release

longshuicy:
Tested publish to the ncsa opensource chart.  You can find clowder2 in the index there: https://opensource.ncsa.illinois.edu/charts/index.yaml

Here is a the successful GH action https://github.com/clowder-framework/clowder2/actions/runs/7790642353/job/21244963744

Added conditional release back so it should only trigger when we tag and release


.github/workflows/helm-release.yaml
```diff
@@ -0,0 +1,73 @@
...
+      - 'main'
+
+    tags:
+     - 'v*'
+
+  pull_request:
+
+  release:
+    types:
+      - 'created'
```
longshuicy, line 19: maybe I should switch this to published

PR #913, opened 2024-02-05
911 public dataset and file page still query private metadata endpoints

tcnichol:
This is now fixed. In public pages, only the public metadata definitions are queried.


frontend/src/components/datasets/PublicDataset.tsx
```diff
@@ -13,20 +13,18 @@ import { useParams, useSearchParams } from "react-router-dom";
...
-
 import { a11yProps, TabPanel } from "../tabs/TabComponent";
 import FilesTable from "../files/FilesTable";
 import { DisplayMetadata } from "../metadata/DisplayMetadata";
 import { DisplayListenerMetadata } from "../metadata/DisplayListenerMetadata";
 import { MainBreadcrumbs } from "../navigation/BreadCrumb";
 import {
 	deleteDatasetMetadata as deleteDatasetMetadataAction,
-	fetchMetadataDefinitions,
+	fetchDatasetMetadata, fetchMetadataDefinitions, fetchPublicMetadataDefinitions,
```
longshuicy, line 27: I don't think you are using the "fetchPublicMetadataDefinitions"? Seems like it's still using the private route. Could you double check if the conflict fix has introduce some change. Thanks!

tcnichol: I got this fixed and committed.

longshuicy (approved):

PR #926, opened 2024-02-12
900 context path

longshuicy:
The frontend "basename" is not at run time. You'll probably build images given a pre-determined basename.


longshuicy:
@max-zilla I recall you made some progress deploying under a subroute. 
If possible, could you share your branch/ example where you've been experimenting with?

PR #928, opened 2024-02-13
Release/v2.0.0 beta2

longshuicy:
rename the branch for the github action to kick in


tcnichol:
I just pushed a small change in DisplayListenerMetadata and now the machine metadata is visible. I moved the key={idx} to the Grid component instead of Card.

I'm not sure why this fixed it, because the other way was in a number of previous commits and I don't remember seeing this bug.

PR #930, opened 2024-02-13
929 message on empty page if there is no datasets

tcnichol:
This is a draft pull request. I don't think it looks very good visually, but it adds a message and a button to register or login on the public page if there are no public datasets.

### DIFF
--- a/frontend/src/components/Public.tsx
+++ b/frontend/src/components/Public.tsx
@@ -101,26 +101,6 @@ export const Public = (): JSX.Element => {
 							) : (
 								<Box>
 									<p>No public datasets on this instance. Login or register to create datasets. </p>
-									<div>
-										<Button
-											component={RouterLink}
-											to="/auth/login"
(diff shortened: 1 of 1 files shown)


ddey2:
I think the login and register button are not needed. We laready have them on top right of the page. Just a message for now is good, I think.

tcnichol:
> I think the login and register button are not needed. We already have them on top right of the page. Just a message for now is good, I think.

I removed them. 

It might be a good idea to maybe have a place where, for each instance, an admin can set a default message. But that would probably be another issue. There was something like this in clowder v1 I recall.

ddey2:
I think this is good for now. We may want to go for a landing page in future. That can be a separate issue.

longshuicy:
For some reason i got 404 error access public_dataset endpoint
image

lmarini:
Please make it against beta.2

longshuicy (approved):

PR #935, opened 2024-02-19
Drag and Drop One or More Files

tcnichol:
A new feature is added. We can now either drag and drop or select multiple files using the same upload dialog. The old ones have been removed since this handles both single and multiple files using either method. 

This currently just works with files, not folders. A zip file is uploaded as a zip file. More options to come with later pull requests.


tcnichol:
Should I make this against the next release instead of main?

ddey2:
I tested, it works. I was able to successfully drag and drop one/multiple files. I have few UI suggestions. 
1. The button already says 'Choose a file', I think the text saying 'Upload a file' may be unnecessary. I was looking for other softwares which supports drag n drop. In Google drive, they don't say anything, but you can simply drag a file anywhere to upload a file. I think we may just want a simple text below the button 'Drop files' to indicate Clowder support the feature.
Screenshot 2024-02-20 at 2 21 49 PM
Screenshot 2024-02-20 at 2 22 25 PM
2. wierd thing ---> I see 'n' after the button. Once I click on the button and uploads a file, it's gone but something else appears.

tcnichol:
@ddey2 
I took out the text from the drag and drop area.

For the n, I haven't been able to replicate it locally. Let me know if it still shows up.

ddey2:
@tcnichol I still see the n. For testing, I opened it in a private Mozilla window, it showed a different label for the button along with a N. I think you're probably using inbuilt class. @longshuicy  Could you check at your end if you see something like the attached ss?
Mozilla_ss
Safari_ss

PR #938, opened 2024-02-20
License model

ddey2:
Made changes on top of this. Please review [this](https://github.com/clowder-framework/clowder2/pull/963) which is pointing this branch.


tcnichol:
For the build issues, were you having the problem where you run pipenv run black app but no files change?

I started with a new repository and environment and I was able to at least fix those. Just pushed. Usually when that has happened locally with me I had to make sure to delete the pipenv environment, since sometimes it would import it rather than create a new one.

lmarini:
Commits included in #963 .

PR #939, opened 2024-02-21
fix the license initialization

longshuicy:
I think i fixed it. 
There are a few places I cleaned up.
But i think the culprit is you have to initialize the LicenseIn correctly, as well as the "holder" part you need to give it a default.
Please check and let me know if you have any questions.

### DIFF
--- a/backend/app/models/licenses.py
+++ b/backend/app/models/licenses.py
@@ -1,10 +1,11 @@
 from datetime import datetime
 from typing import List
 
+from beanie import Document
 from pydantic import BaseModel, Field
-from beanie import Document, View, PydanticObjectId
 
-from backend.app.models.users import UserOut
+from app.models.authorization import Provenance
+from app.models.users import UserOut
 
 
 class LicenseBase(BaseModel):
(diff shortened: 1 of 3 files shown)


backend/app/models/licenses.py
```diff
@@ -13,20 +14,18 @@ class LicenseBase(BaseModel):
     text: str
     url: str
     version: str
-    holders: List[UserOut]
-    expiration_date: datetime = Field(default_factory=datetime.utcnow)
+    holders: List[UserOut] = []
+    expiration_date: datetime = Field(
+        default_factory=datetime.utcnow
+    )  # TODO: shoudn't the default be never?
```
ddey2, line 20: Yes, will change that. It was just a draft.

backend/app/models/licenses.py
```diff
@@ -13,20 +14,18 @@ class LicenseBase(BaseModel):
...
 class LicenseIn(LicenseBase):
     pass
 
 
-class LicenseDB(Document, LicenseBase):
-    creator: UserOut
-    created: datetime = Field(default_factory=datetime.utcnow)
-    modified: datetime = Field(default_factory=datetime.utcnow)
-
+class LicenseDB(Document, LicenseBase, Provenance):
```
ddey2, line 28: I understand that Provenance gets the metadata, but was there error in the older version?

backend/app/models/licenses.py
```diff
@@ -13,20 +14,18 @@ class LicenseBase(BaseModel):
     text: str
     url: str
     version: str
-    holders: List[UserOut]
-    expiration_date: datetime = Field(default_factory=datetime.utcnow)
+    holders: List[UserOut] = []
```
ddey2, line 17: I think this is the key change. I wonder why this creates issue in initializing the class, when no other variables are initialized.

ddey2 (approved):

PR #951, opened 2024-02-29
enable superadmin does not update privileges

tcnichol:
I added adminMode as a boolean in Authwrapper itself, and now the dataset (and files inside the dataset) update their menus as I enable superadmin mode and they go away as soon as I disable it.

This is draft because I want to do more testing to make sure that every place where permissions should update work correctly.


longshuicy (approved):
works well

PR #952, opened 2024-02-29
fix button

longshuicy:
image


tcnichol (approved):
Checked and this is fixed. Thanks

PR #964, opened 2024-03-08
836 public search option

tcnichol:
This pull request will enable a search of public data for non-logged in users.

It also fixes a problem. When a logged in user would search, public and authenticated matches were not being shown. In the query 'public' and 'authenticated' had to be in all lowercase. Additionally, not all files were having their status updated when the dataset changed. This is also fixed.

To test, create datasets. Make sure some are public and authenticated.

If no user is logged in, public datasets and files should show up in search results. 

If a user is logged in, public and authenticated datasets and files should show up in search results, along with their own items or datasets explicitly shared. 

Changing the status of a dataset should be reflected in search results.


ddey2 (changes requested):
@tcnichol I can see the public datasets in the dropdown list for public search. But when I click enter or click on any of the dataset form the dropdown list, it doesn't take me to the search result page or the dataset page itself.
Screenshot 2024-03-11 at 3 31 35 PM

tcnichol:
> @tcnichol I can see the public datasets in the dropdown list for public search. But when I click enter or click on any of the dataset form the dropdown list, it doesn't take me to the search result page or the dataset page itself. Screenshot 2024-03-11 at 3 31 35 PM

I think I have fixed this. I noticed that in the search results the links were not for public_datasets or public_files, they were the same as for logged in search. I added a PublicSearchResult component which should fix these issues.

frontend/src/components/search/PublicSearch.tsx
```diff
@@ -0,0 +1,210 @@
...
+									onChange={() => {
+										setLuceneOn((prevState) => !prevState);
+									}}
+									name="Query String"
+								/>
+							}
+							label={
+								<Typography
+									variant="body1"
+									sx={{
```
ddey2, line 47: Can you put the styles in main.css, may be?

frontend/src/components/search/PublicSearchResult.tsx
```diff
@@ -0,0 +1,104 @@
...
+}
+
+export function PublicSearchResult(props) {
+	const { data } = props;
+
+	if (data.length > 0) {
+		return (
+			<List
+				sx={{
+					width: "100%",
```
ddey2, line 87: here too

frontend/src/components/search/PublicSearchResult.tsx
```diff
@@ -0,0 +1,104 @@
...
+	return (
+		<>
+			<ListItemAvatar sx={{ color: theme.palette.primary.main }}>
+				<InsertDriveFileIcon />
+			</ListItemAvatar>
+			<Box sx={{ marginTop: "5px" }}>
+				<MuiLink
+					component={Link}
+					to={`/public_files/${item._id}?public_dataset=${item.dataset_id}`}
+					sx={{ fontWeight: "bold", fontSize: "18px" }}
```
ddey2, line 65: here too.

frontend/src/components/search/PublicSearchResult.tsx
```diff
@@ -0,0 +1,104 @@
...
+	return (
+		<>
+			<ListItemAvatar sx={{ color: theme.palette.primary.main }}>
+				<DatasetIcon />
+			</ListItemAvatar>
+			<Box sx={{ marginTop: "5px" }}>
+				<MuiLink
+					component={Link}
+					to={`/public_datasets/${item._id}`}
+					sx={{ fontWeight: "bold", fontSize: "18px" }}
```
ddey2, line 40: here too...looks like you're using the same styling properties again.

ddey2 (changes requested):

frontend/src/components/search/PublicSearchResult.tsx
tcnichol: I created a separate issue for these.

tcnichol:
@longshuicy 

I checked and the search result links should be fixed for public search now, for both files and datasets.

frontend/src/components/PublicLayout.tsx
```diff
@@ -235,6 +235,18 @@ export default function PersistentDrawerLeft(props) {
 						</ListItemButton>
 					</ListItem>
 				</List>
+				<Divider/>
+				<List>
+					<ListItem key={"public_search"} disablePadding>
+						<ListItemButton component={RouterLink} to="/public_search">
+							<ListItemIcon>
+								<SearchDatasetIcon />
```
ddey2, line 243: It gives error here as you never imported SearchDatasetIcon in this file.

ddey2 (changes requested):
Rest looks good to me. Once you import SearchDatasetIcon, public search works.

frontend/src/components/PublicLayout.tsx
longshuicy: @tcnichol is this addressed?

tcnichol: Just pushed a fix for this.

ddey2:
@tcnichol Clicking on the link after the search doesn't work. Clicking on the dataset takes you to the main page.
Screenshot 2024-04-16 at 2 39 53 PM

ddey2 (changes requested):
left comment

tcnichol:
@ddey2 I noticed that there was an inconsistency in the routes for public_datasets. Some spots had public/datasets instead. Went through and I think I fixed that.

Though now the link to the public search is broken at the top of the Public page, working on that.

ddey2 (approved):
Looks good.

backend/app/routers/elasticsearch.py
```diff
@@ -23,8 +23,8 @@ def _add_permissions_clause(
             "should": [
                 {"term": {"creator": username}},
                 {"term": {"user_ids": username}},
-                {"term": {"status": "AUTHENTICATED"}},
-                {"term": {"status": "PUBLIC"}},
+                {"term": {"status": "authenticated"}},
+                {"term": {"status": "public"}},
```
longshuicy, line 27: Just curious, why capitalized doesn't work? In the database and in the router logic, it's all capitalized?

frontend/src/routes.tsx
```diff
@@ -194,7 +195,7 @@ export const AppRoutes = (): JSX.Element => {
 						</PrivateRoute>
 					}
 				/>
-				<Route path="/public/files/:fileId" element={<PublicFileComponent />} />
+				<Route path="/public_files/:fileId" element={<PublicFileComponent />} />
```
longshuicy, line 198: Are we using public/files or public_files? 
In other PRs that got merged in, I fixed places that are using "_"...

longshuicy: If you would like to use "_", you might have to 1. merge the latest main, 2. do a project wide search of places still using /
E.g. If you go to a file under public dataset, it right now still goes to /public/file and hence broken.

longshuicy (changes requested):
Public search seems to be able to match dataset, not files though. Might be your clause is specifically matching dataset status. Not sure how to get around, but at least I would change the label `Search for Datasets and Files`
image

frontend/src/routes.tsx
tcnichol: I think I got all of these fixed with a merge and a search/replace. Testing.

tcnichol: I also think that files should have a status. Checking the populate fake data, the status for files is 'null' in some cases and those should match the datasets.

backend/app/routers/elasticsearch.py
tcnichol: I have fixed the problem with the files having null status. They now should always have the same status as the dataset.

I'm still not getting files in the search results for public search, looking into this.

frontend/src/routes.tsx
longshuicy: Thank you. This has been resolved.

backend/app/routers/elasticsearch.py
longshuicy: I don't think this has been resolved yet?

longshuicy (changes requested):

backend/app/routers/elasticsearch.py
tcnichol: I think this is fixed. Files should be showing up in public search now.

longshuicy:
@tcnichol any chance you have addressed public files not searchable issue? If it is challenging, could you create another issue to address that separately? The rest of this PR looks good.

tcnichol:
> @tcnichol any chance you have addressed public files not searchable issue? If it is challenging, could you create another issue to address that separately? The rest of this PR looks good.

Let me check. I think it should be fixed because the problem was that the fields for public or authenticated were null for files in elasticsearch, but that should be fixed now. I'll make sure it is working.

lmarini:
> > @tcnichol any chance you have addressed public files not searchable issue? If it is challenging, could you create another issue to address that separately? The rest of this PR looks good.
> 
> Let me check. I think it should be fixed because the problem was that the fields for public or authenticated were null for files in elasticsearch, but that should be fixed now. I'll make sure it is working.

@tcnichol is this fixed now? Thanks.

longshuicy (approved):
After drop volume and restart, elasticsearch is indexing properly and I can search for files. Approved.

PR #965, opened 2024-03-11
add my datasets tab

tcnichol:
This adds 'my datasets' so that a user can see just their datasets on another tab, separate from the shared, authenticated and public datasets.


tcnichol:
all datasets
listdatasets


and 'my datasets' tab

mydatasets

longshuicy (changes requested):
Another thing is SuperAdmin doesn't immediately reflected on the MyDataset page. If you refresh the page then you see it. Need to add superAdmin in the useEffect hook.

frontend/src/components/Explore.tsx
```diff
@@ -49,6 +56,11 @@ export const Explore = (): JSX.Element => {
 		listDatasets(0, limit, mine);
 	}, []);
 
+	// component did mount
+	useEffect(() => {
+		listMyDatasets(0, limit);
```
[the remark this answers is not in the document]
tcnichol: I think I fixed this. For admin mode, if you are sending in 'mine' it only looks at datasets from the user, and ignores admin mode.

tcnichol:
I believe I have fixed some pagination issues with myDatasets. I needed some extra methods so that pagination is separated for each tab.

frontend/src/actions/dataset.js
```diff
@@ -220,6 +220,25 @@ export function fetchDatasets(skip = 0, limit = 21, mine = false) {
 	};
 }
 
+export const RECEIVE_MY_DATASETS = "RECEIVE_MY_DATASETS";
+
+export function fetchMyDatasets(skip = 0, limit = 21, mine = true) {
+	return (dispatch) => {
+		// TODO: Parameters for dates? paging?
+		return V2.DatasetsService.getDatasetsApiV2DatasetsGet(skip, limit, true)
```
longshuicy, line 228: Any reason you can reuse the fetch Datasets? I see you are hardcoding "mine" to be true in the function. Wondering if it's the same effect as pass in mine=true as parameter

tcnichol: I had added myDatasets since I added a new field in the root state, so it has a new method that sets it. 
I can try to just use the same state though, looking at it now I think that it should work fine with just 1 state.

backend/app/routers/datasets.py
```diff
@@ -228,7 +228,7 @@ async def get_datasets(
     admin=Depends(get_admin),
     admin_mode: bool = Depends(get_admin_mode),
 ):
-    if admin and admin_mode:
+    if admin and admin_mode and not mine:
```
longshuicy, line 231: Just double check, based on this logic, we want to completely separate out other people's dataset vs "mine" dataset?

frontend/src/components/Explore.tsx
```diff
@@ -49,16 +52,30 @@ export const Explore = (): JSX.Element => {
...
+
+
 	// switch tabs
 	const handleTabChange = (
 		_event: React.ChangeEvent<{}>,
 		newTabIndex: number
 	) => {
+		if (newTabIndex === 1) {
+			setMine(true);
+			listDatasets(0, limit, true);
```
longshuicy, line 77: I don't think you need this `listDatasets`. You have a useEffect hook listen to Mine

backend/app/routers/datasets.py
tcnichol: Yes. I was thinking that if we enable superadmin, it wouldn't be helpful to have both tabs be the same and just contain all the datasets.

longshuicy (changes requested):

frontend/src/components/Explore.tsx
```diff
@@ -49,16 +52,30 @@ export const Explore = (): JSX.Element => {
...
+
 	// Admin mode will fetch all datasets
 	useEffect(() => {
 		listDatasets(0, limit, mine);
 	}, [adminMode]);
 
+	// Admin mode will fetch my datasets
+	useEffect(() => {
+		listDatasets(0, limit, true);
+	}, [adminMode, mine]);
```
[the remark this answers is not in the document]
tcnichol: This is combined now

frontend/src/components/Explore.tsx
tcnichol: This is removed, not needed

frontend/src/components/Explore.tsx
```diff
@@ -34,13 +34,16 @@ export const Explore = (): JSX.Element => {
 	const pageMetadata = useSelector(
 		(state: RootState) => state.dataset.datasets.metadata
 	);
+	const pageMyDatasetsMetadata = useSelector(
+		(state: RootState) => state.dataset.myDatasets.metadata
```
ddey2, line 38: I don't see that you added myDatasets under RootState.

It doesn't work as it find myDatasets as defined. Declare it under RootState in data.ts

ddey2 (changes requested):
doesn't work

frontend/src/components/Explore.tsx
tcnichol: I added the state and new actions and methods. Should be working now.

ddey2 (approved):

lmarini (changes requested):
If I flip to My Datasets and then go back to all datasets I only see my datasets.

tcnichol:
> If I flip to My Datasets and then go back to all datasets I only see my datasets.

When the tab is changed, I wasn't changing 'mine' after the tab was clicked back. This should be fixed now.

frontend/src/components/Explore.tsx
```diff
@@ -49,16 +52,33 @@ export const Explore = (): JSX.Element => {
...
+
 	// switch tabs
 	const handleTabChange = (
 		_event: React.ChangeEvent<{}>,
 		newTabIndex: number
 	) => {
+		if (newTabIndex === 1) {
+			setMine(true);
+			listDatasets(0, limit, true);
+		} else if (newTabIndex === 0) {
```
longshuicy, line 78: This logic might be more robust if you just say "else". Because only when tab index is 1, it is "mine" tab. Others should all set Mine to "false".

longshuicy (commented):
Pagination is still not quite right. To reproduce:
1. on the all dataset page, flip a few pages
2. switch to "my" tab
3. the previous button becomes clickable even though I just have one page. Upon clicking the flip button, I get an empty page

image
image

tcnichol:
> Pagination is still not quite right. To reproduce:
> 
> 1. on the all dataset page, flip a few pages
> 2. switch to "my" tab
> 3. the previous button becomes clickable even though I just have one page. Upon clicking the flip button, I get an empty page
> 
> image image

I think this is fixed now. I can go to a later page, switch to 'my datasets' and then I'll be back at 1, with only 1 page showing.

longshuicy:
> > Pagination is still not quite right. To reproduce:
> > 
> > 1. on the all dataset page, flip a few pages
> > 2. switch to "my" tab
> > 3. the previous button becomes clickable even though I just have one page. Upon clicking the flip button, I get an empty page
> > 
> > image image
> 
> I think this is fixed now. I can go to a later page, switch to 'my datasets' and then I'll be back at 1, with only 1 page showing.

But somehow the previous button is still clickable and I'm not sure it's quite on 1
image

lmarini:
> > > Pagination is still not quite right. To reproduce:
> > > 
> > > 1. on the all dataset page, flip a few pages
> > > 2. switch to "my" tab
> > > 3. the previous button becomes clickable even though I just have one page. Upon clicking the flip button, I get an empty page
> > > 
> > > image image
> > 
> > 
> > I think this is fixed now. I can go to a later page, switch to 'my datasets' and then I'll be back at 1, with only 1 page showing.
> 
> But somehow the previous button is still clickable and I'm not sure it's quite on 1 image

@tcnichol is this fixed now? Thanks.

tcnichol:
> > > > Pagination is still not quite right. To reproduce:
> > > > 
> > > > 1. on the all dataset page, flip a few pages
> > > > 2. switch to "my" tab
> > > > 3. the previous button becomes clickable even though I just have one page. Upon clicking the flip button, I get an empty page
> > > > 
> > > > image image
> > > 
> > > 
> > > I think this is fixed now. I can go to a later page, switch to 'my datasets' and then I'll be back at 1, with only 1 page showing.
> > 
> > 
> > But somehow the previous button is still clickable and I'm not sure it's quite on 1 image
> 
> @tcnichol is this fixed now? Thanks.

I have checked and now when you click 'my datasets' the 'previous' button is no longer active.

longshuicy:
The above mentioned pagination issue has resolved;
However I discovered another one:

1. If you have many pages on "My dataset" tab
2. Flip to a later page
3. Switch back to "Dataset" tab
4. It will be flipped to that page. Maybe we would like to start from the first page to be consistent?

image

tcnichol:
> The above mentioned pagination issue has resolved; However I discovered another one:
> 
> 1. If you have many pages on "My dataset" tab
> 2. Flip to a later page
> 3. Switch back to "Dataset" tab
> 4. It will be flipped to that page. Maybe we would like to start from the first page to be consistent?
> 
> image

I see that I forgot to change page when tab switching to 'datasets.' Now it should behave the same as switching to my datasets - you will go back to the first page.

PR #967, opened 2024-03-12
style of the sharing table; link to the group/{group_id} page

longshuicy:
- Clean up the sharing sub table
- Link group to the /groups/{id} 

image


tcnichol:
This looks good. I just see an error for codegen, but aside from that this is good to go.

tcnichol (approved):
Since codegen is now fixed I will mark this approved.

PR #968, opened 2024-03-12
clear more breakpoints and move role to the Details

longshuicy:
- Breakpoint should work ok now
- Move dataset description to the correct container
- Move RoleChip to the Details for both file and dataset

image


tcnichol (approved):
This looks good. Details changes on all pages, marking approved.

ddey2 (approved):
looks good

PR #972, opened 2024-03-13
remove analysis tab from authenticated file

tcnichol:
Removing the Analysis tab from authenticated files. The user should not have access to that tab, matching behavior with datasts.


frontend/src/components/files/File.tsx
```diff
@@ -115,7 +118,7 @@ export const File = (): JSX.Element => {
 	// snack bar
 	const [snackBarOpen, setSnackBarOpen] = useState(false);
 	const [snackBarMessage, setSnackBarMessage] = useState("");
-
+	console.log('file role', fileRole);
```
longshuicy, line 121: could you remove the console.log?

frontend/src/components/files/File.tsx
```diff
@@ -343,14 +346,18 @@ export const File = (): JSX.Element => {
...
 						/>
-						<Tab
-							icon={<BuildIcon />}
-							iconPosition="start"
-							sx={TabStyle}
-							label="Analysis"
-							{...a11yProps(3)}
-							disabled={false}
-						/>
+						{datasetRole.role !== undefined && datasetRole.role !== "viewer" ? (
```
longshuicy, line 349: Could we use the AuthWrapper here?

frontend/src/components/files/File.tsx
```diff
@@ -435,9 +442,13 @@ export const File = (): JSX.Element => {
 							version={fileSummary.version_num}
 						/>
 					</TabPanel>
-					<TabPanel value={selectedTabIndex} index={3}>
-						<Listeners fileId={fileId} datasetId={datasetId} />
-					</TabPanel>
+					{datasetRole.role !== undefined && datasetRole.role !== "viewer" ? (
```
longshuicy, line 445: Same could we use the "AuthWrapper" component here to wrap around?

longshuicy (changes requested):

frontend/src/components/files/File.tsx
tcnichol: removed

frontend/src/components/files/File.tsx
tcnichol: changed to authwrapper

frontend/src/components/files/File.tsx
tcnichol: changed here too

longshuicy (approved):
Works.

tcnichol:
Merged beta into this branch.

PR #973, opened 2024-03-18
Keycloak enabled change

bingzhang:
1. use clowder admin to login to get bearer token.
http://localhost:8000/api/v2/login

2.  disable existing user email account (e.g., bing2@illinois.edu).
http://localhost:8000/api/v2/users/disable/bing2@illinois.edu

test the login from clowder, if you cannot login then that is a correct behavior.

3. enable it.
http://localhost:8000/api/v2/users/enable/bing2@illinois.edu

test the login from clowder, if you can login then that is a correct behavior.


longshuicy (commented):
This is really cool. 
Just curious, if you enable a user, in the keycloak interface how can I see this get reflected?
Also if you have time, could yo include a quick pytest. Test is under "backend/app/tests" folder. Thanks!

longshuicy (approved):
@bingzhang your demo answered all of my question. Thanks!

max-zilla (approved):

max-zilla:
looks good, everything seemed to work as expected

backend/app/routers/authentication.py
```diff
@@ -201,3 +201,67 @@ async def revoke_admin(
...
+                )
+    else:
+        raise HTTPException(
+            status_code=403,
+            detail=f"User {current_username.email} is not an admin. Only admin can enable user access.",
+        )
+
+
+@router.post("/users/disable/{useremail}", response_model=UserOut)
+async def user_enable(
```
ddey2, line 239: Do you mind renaming this function to user_disable? Same function names might be confusing and this function is actually disabling users. 
 Rest look all good to me.

ddey2 (changes requested):

backend/app/routers/authentication.py
bingzhang: fixed it. Thank you.

ddey2 (approved):
Looks good.

lmarini:
@bingzhang can you resolve conflict? Thanks!

longshuicy:
Tested it again, works as expected.

PR #974, opened 2024-03-21
866 readonly user

tcnichol:
The idea here is to add a read only user.

To test, create users and then modify one of the users in MongoDB to have 'read_only_user' == true. 

The idea is that if a user is read only, that will limit them to only being a Viewer. 
Intended behavior for groups (not implemented) - if a read only user is added to a group, that user will only be a Viewer for datasets  regardless of group level permission.

Problem right now. After trying to share with a user, I want to go through the datasetRoles and see if the new user was successfully added. I think there is something wrong with my hook because after I add the user, even when I call getRoles(datasetId), the new user is not added but it is being returned by the back end endpoint.


tcnichol:
This is draft, but I am having a problem with the MangeUsers component.

Toggling the switch for read only user does change the value of the boolean for the user, but this change is not reflected unless you refresh the page.

tcnichol:
I am marking this pull request ready for review.

How the read only user works, at least for now. 

A user can be made a read only user by an admin in the "Manage Users"  component. A read only user can only have Viewer permissions for any dataset in the instance. Trying to add a read only user as other roles will throw an error. 

For groups. A read only user can be added to a group, but no matter what role the group is assigned, that user will still only be a Viewer. So if the group has Editor access, the read only user in the group will only be a Viewer. 

This seemed like the least confusing way for users to share data with a read only user. They might share data with a particular group, and can just add the read only user. This seemed to be less complicated than having to create another group or share each dataset individually. 

A read only user can also not be made an admin. This seemed to make sense since if the user was admin, they would have all permissions.

backend/app/keycloak_auth.py
```diff
@@ -306,6 +306,91 @@ async def get_current_username(
     )
 
 
+async def get_read_only_user(
```
ddey2, line 309: I have a query. I get that this function is returning is an user is a read only user. I wonder about how 
1. getting the mongo object from get_current_user
2. then getting the read_only_field from that mongo object?

Won't this make the function simpler? or am I missing somthing here?

backend/app/routers/authentication.py
```diff
@@ -199,3 +202,39 @@ async def revoke_admin(
...
         )
+
+
+@router.post("/users/enable_readonly/{useremail}", response_model=UserOut)
+async def enable_readonly_user(
+    useremail: str, current_username=Depends(get_current_user), admin=Depends(get_admin)
+):
+    if admin and current_username.admin:
+        if (user := await UserDB.find_one(UserDB.email == useremail)) is not None:
+            user.read_only_user = True
```
ddey2, line 213: I was wondering if we can create a helper function to enable/disable users since the 2 functions is almost exactly same apart from setting
user.read_only_user = True/False
and use the helper function.

ddey2 (commented):
Regarding the use cases, I have a query - 
I can set a user as read_only_user who is an admin. Should we actually do that? Won't admin means that the user should have all access?
On turning off admin, I can't make the user as admin again, as it is a read_only_user. This makes sense, I guess. So, in 1st scenario too, I think we won't let that.
Screenshot 2024-04-11 at 10 40 58 AM
Screenshot 2024-04-11 at 10 41 57 AM

backend/app/routers/authentication.py
tcnichol: I tried doing this, but somehow the new function never got called by the front end after I had added it.

This might be related to codegen not working. I had to add it into the services manually, so maybe that is why it does not work.

ddey2 (changes requested):
I see several unresolved modules errors.

lmarini (changes requested):
Dataset and files look good, but I was able to create a new dataset and add a metadata definition.

lmarini:
> I see several unresolved modules errors.

@ddey2 do you still have module errors?

tcnichol:
> Regarding the use cases, I have a query - I can set a user as read_only_user who is an admin. Should we actually do that? Won't admin means that the user should have all access? On turning off admin, I can't make the user as admin again, as it is a read_only_user. This makes sense, I guess. So, in 1st scenario too, I think we won't let that. Screenshot 2024-04-11 at 10 40 58 AM Screenshot 2024-04-11 at 10 41 57 AM

I changed this and now you cannot make an admin a read only user. It will be disabled there.

ddey2:
> > I see several unresolved modules errors.
> 
> @ddey2 do you still have module errors?

those errors are fixed now.

ddey2 (changes requested):
@tcnichol However, I see that I can't turn off the 'Read only' toggle switch even for non-admin users.  In the below ss, clicking on read-only switch doesn't work.
Screenshot 2024-05-13 at 4 00 07 PM

tcnichol:
> @tcnichol However, I see that I can't turn off the 'Read only' toggle switch even for non-admin users. In the below ss, clicking on read-only switch doesn't work. Screenshot 2024-05-13 at 4 00 07 PM

let me check this out.

tcnichol:
For now, 

I removed 'groups' and 'new dataset' from the left drawer for the read only user.
I also allowed the read only user to view metadata definitions but not create or delete.

Will need to check other API routes to make sure read only user cannot create using the API.

backend/app/routers/authentication.py
```diff
@@ -206,6 +206,42 @@ async def revoke_admin(
         )
 
 
+@router.post("/users/enable_readonly/{useremail}", response_model=UserOut)
+async def enable_readonly_user(
+    useremail: str, current_username=Depends(get_current_user), admin=Depends(get_admin)
+):
+    if admin and current_username.admin:
```
longshuicy, line 213: Should this be `if admin and current_username.admin_mode:`? Otherwise you are checking the same thing twice.

backend/app/routers/authentication.py
```diff
@@ -206,6 +206,42 @@ async def revoke_admin(
...
+        if (user := await UserDB.find_one(UserDB.email == useremail)) is not None:
+            user.read_only_user = True
```
longshuicy, line 215: you might not want to make an admin user "read_only_user" and vice versa?

backend/app/routers/authentication.py
```diff
@@ -206,6 +206,42 @@ async def revoke_admin(
...
+            status_code=403,
+            detail=f"User {current_username.email} is not an admin. Only admin can make others admin.",
+        )
+
+
+@router.post("/users/disable_readonly/{useremail}", response_model=UserOut)
+async def disable_readonly_user(
+    useremail: str, current_username=Depends(get_current_user), admin=Depends(get_admin)
+):
+    if admin and current_username.admin:
```
longshuicy, line 231: Same as above

frontend/src/components/Layout.tsx
```diff
@@ -154,7 +154,7 @@ export default function PersistentDrawerLeft(props) {
 	}, [location]);
 
 	const loggedOut = useSelector((state: RootState) => state.error.loggedOut);
-
+	console.log("user profile", currUserProfile);
```
longshuicy, line 157: console.log

frontend/src/components/metadata/MetadataDefinitions.tsx
```diff
@@ -80,6 +83,11 @@ export function MetadataDefinitions() {
 		listMetadataDefinitions(null, 0, limit);
 	}, []);
 
+	// get profile
+	useEffect(() => {
+		fetchCurrUserProfile();
+	}, []);
```
longshuicy, line 89: you can combine this with the above useEffect

longshuicy (commented):
Once I turned on read only, i cannot turn it off. Seems like a frontend problem because I see no interaction with backend.

image

--
Once I turned on read-only first, then turn on admin; read-only seems to be frozen.
image

backend/app/routers/authentication.py
tcnichol: fixed. cannot make an admin user read only

longshuicy:
> Once I turned on read only, i cannot turn it off. Seems like a frontend problem because I see no interaction with backend.
> 
> image
> Once I turned on read-only first, then turn on admin; read-only seems to be frozen. image

The first has been solved. Not sure if the "Once I turned on read-only first, then turn on admin; read-only seems to be frozen." is by design.

longshuicy (approved):
Features works now. Could you resolve merge conflict? Thanks!

lmarini (approved):

PR #977, opened 2024-03-28
Implement edit metadata definition 

ddey2:
Steps to test:
1. create a metadata definition
2. Click on it
3. Click on edit button on top right
4. Play around with various fields of metadata definition and 
Screenshot 2024-04-11 at 10 02 41 PM
Screenshot 2024-04-11 at 10 02 47 PM
try to update it.


tcnichol:
Something that I tried.

I created a metadata definition where the input type was a text string.

Then I changed it to a date picker. 

When I go to the file with the metadata, and I edit it, I get the new select.

But if I click on the 'add metadata' button instead of editing the existing metadata, I get a blank screen.

tcnichol:
I also noticed that the 'Add Metadata' button is missing on the Dataset page on this branch, but it is present in main. The Dataset view isn't changed.

ddey2:
> I also noticed that the 'Add Metadata' button is missing on the Dataset page on this branch, but it is present in main. So this branch might just need a merge with main to fix that.

merged main

ddey2:
> Something that I tried.
> 
> I created a metadata definition where the input type was a text string.
> 
> Then I changed it to a date picker.
> 
> When I go to the file with the metadata, and I edit it, I get the new select.
> 
> But if I click on the 'add metadata' button instead of editing the existing metadata, I get a blank screen. This happens both before I edit the metadata value (when it's still a string even though the definition is a DateTimePicker) or after.

looks like "Add metadata" button is not working on main. I created a separate issue #1033

frontend/src/reducers/metadata.ts
```diff
@@ -67,6 +68,11 @@ const metadata = (state = defaultState, action: DataAction) => {
 			return Object.assign({}, state, {
 				newMetadataDefinition: action.metadataDefinition,
 			});
+		case EDIT_METADATA_DEFINITION:
+			console.log("here");
```
Vismayak, line 72: We can remove the log statement 😄

frontend/src/components/metadata/MetadataDefinitionEntry.tsx
```diff
@@ -131,6 +157,17 @@ export function MetadataDefinitionEntry() {
...
+						variant="contained"
+						aria-label="edit"
+						onClick={() => {
+							setEditMetadataDefinitionOpen(true);
+						}}
+						endIcon={<EditIcon />}
+						sx={{ float: "right" }}
+					>
+						Edit
+					</Button>
```
longshuicy, line 170: image

Could you add a little bit margin between these two buttons?

frontend/src/components/metadata/EditMetadataDefinitionModal.tsx
```diff
@@ -0,0 +1,671 @@
...
+											{idx == 0 ? (
+												<></>
+											) : (
+												<IconButton
+													color="primary"
+													size="small"
+													onClick={() => removeContext(idx)}
+												>
+													<DeleteOutlineIcon />
+												</IconButton>
```
longshuicy, line 432: image

The first item cannot be deleted in this case. You might want to relax the `idx == 0 ?` criterion

frontend/src/components/metadata/EditMetadataDefinitionModal.tsx
```diff
@@ -0,0 +1,671 @@
...
+								) : (
+									<StepButton
+										color="inherit"
+										onClick={() => {
+											setActiveStep(idx + 1);
+										}}
+									>
+										<StepLabel>
+											Add additional entry
+											{idx == activeStep - 1 ? (
```
longshuicy, line 482: Once click add metadata entry and save updated; the next time when editing, it is still showing "add additional entry"
image

frontend/src/components/metadata/EditMetadataDefinitionModal.tsx
```diff
@@ -0,0 +1,671 @@
...
+							<Step key={idx}>
+								{idx == 0 ? (
+									<StepButton
+										color="inherit"
+										onClick={() => {
+											setActiveStep(idx + 1);
+										}}
+									>
+										<StepLabel>
+											Edit metadata entry*
```
longshuicy, line 459: There is no delete entry option right next to the edit metadata
image

longshuicy (changes requested):
- Could you remove all the `console.log`?
- If the metadata definition is modified, but there are already metadata created using this definition, there might be some inconsistency. We might want to think about what should be the behavior

backend/app/routers/metadata.py
```diff
@@ -70,6 +71,49 @@ async def get_metadata_definition_list(
...
+)
+async def update_metadata_definition(
+    metadata_definition: MetadataDefinitionIn,
+    metadata_definition_id: str,
+    user=Depends(get_current_user),
+):
+    existing_count = await MetadataDefinitionDB.find(
+        {
+            "$and": [
+                {"_id": {"$ne": PydanticObjectId(metadata_definition_id)}},
```
longshuicy, line 85: why would "$ne" needed?

backend/app/routers/metadata.py
```diff
@@ -70,6 +71,49 @@ async def get_metadata_definition_list(
...
+                {"name": {"$eq": metadata_definition.name}},
+            ]
+        }
+    ).count()
+    if existing_count > 0:
+        raise HTTPException(
+            status_code=409,
+            detail=f"Metadata definition named {metadata_definition.name} already exists.",
```
longshuicy, line 93: Shouldn't this be the opposite? We would want to update the existing metadata.

backend/app/routers/metadata.py
```diff
@@ -70,6 +71,49 @@ async def get_metadata_definition_list(
...
+                PydanticObjectId(metadata_definition_id)
+            )
+        ) is not None:
+            mdd.name = metadata_definition.name
+            mdd.description = metadata_definition.description
+            mdd.required_for_items = metadata_definition.required_for_items
+            mdd.context = metadata_definition.context
+            mdd.context_url = metadata_definition.context_url
+            mdd.fields = metadata_definition.fields
+            mdd.modified = datetime.datetime.utcnow()
```
longshuicy, line 107: Could this logic be more versatile that we just overwrite everything with metadata_definition without specifically listing each field (except the modified)

frontend/src/reducers/metadata.ts
ddey2: done

frontend/src/components/metadata/MetadataDefinitionEntry.tsx
ddey2: done.

backend/app/routers/metadata.py
ddey2: Want to find out other metadata definitions using the same name.

frontend/src/components/metadata/EditMetadataDefinitionModal.tsx
ddey2: I was thinking users may want to 'add additional entry' in addition to updating the existing ones

backend/app/routers/metadata.py
ddey2: This is for the scenario where the users update the name and the updated name is same as some other existing metadata definition name

ddey2: My understanding is - we want to keep the metadata definiton name as unique. Metadata definition name is used to connect definition to actual metadata.

frontend/src/components/metadata/EditMetadataDefinitionModal.tsx
ddey2: Done.

frontend/src/components/metadata/EditMetadataDefinitionModal.tsx
ddey2: My bad. I got what you meant. Resolved.  For simplicity, I just put 'Edit metadata entry' for now. Deleting metadata entry in btw the list and then tracking each of them seemed to me overkill for now. If needed, I can address that.

frontend/src/components/metadata/EditMetadataDefinitionModal.tsx
ddey2: Done

backend/app/routers/metadata.py
ddey2: I have replaced this block by using setattr.

ddey2:
Right now, I have disabled editing metadata definition if metadata exists adhering to that definition. It will throw exception
Screenshot 2024-06-06 at 10 07 33 PM

frontend/src/components/metadata/EditMetadataDefinitionModal.tsx
longshuicy: looks correct now.

backend/app/routers/metadata.py
longshuicy: Did you actually pushed the setattr part?

backend/app/routers/metadata.py
```diff
@@ -70,6 +71,62 @@ async def get_metadata_definition_list(
...
+                    detail=f"Metadata definition: {mdd.name} ({metadata_definition_id}) in use. "
+                    f"You cannot edit it if there are existing metadata using this definition.",
+                )
+            try:
+                metadata_update = metadata_definition.dict()
+                metadata_update["modified"] = datetime.datetime.utcnow()
+
+                # Update mdd's attributes using the updated dictionary
+                for key, value in metadata_update.items():
+                    setattr(mdd, key, value)
```
ddey2, line 118: @longshuicy here it is

longshuicy (approved):
works well. Thanks!

Vismayak:
works good, doesnt let me edit as wanted if a metadata already exists 😄

Vismayak (approved):

PR #988, opened 2024-04-02
981 clowder2 register link on the topbar doesnt work

longshuicy:
For some reason we are missing the redirect_uri. Maybe it's related to keycloak upgrade?


tcnichol (approved):
Checked and registration works fine for new users now. Marking this one approved.

PR #1001, opened 2024-04-12
996 detail information update on

longshuicy:
Add information tooltip for some of the detail fields. 
- If "info" field available, will display the "i" icon and hover over will show some prompt
- Otherwise stays the same 
- Some of the wording choice might subject to fine-tuning

e.g.

image
image
image


lmarini (approved):
Simplified details keys.

PR #1005, opened 2024-04-15
Docs to set up Clowder in VM 

Vismayak:


longshuicy (approved):

lmarini (changes requested):
Can we rename this to "Standalone docker setup" or something similar since it's not just for VMs, but also other hosts? Then rework the getting of IP address for the host (instead of VM)?

lmarini (approved):

PR #1006, opened 2024-04-15
tentatively set redirect uri to be home for registration

longshuicy:


backend/app/routers/datasets.py
```diff
@@ -452,8 +407,10 @@ async def delete_dataset(
         await AuthorizationDB.find(
             AuthorizationDB.dataset_id == PydanticObjectId(dataset_id)
         ).delete()
-        return {"deleted": dataset_id}
         await delete_license(dataset.license_id)
+
+        return {"deleted": dataset_id}
```
longshuicy, line 412: Here is a bug I fixed. Before you have @ddey2 
-  await delete_license after the return which makes this code never reachable.
- didn't import delete_license properly

ddey2 (approved):
works now.

PR #1008, opened 2024-04-16
Delete dataset will also delete custom license

longshuicy:
Original code has wrong order or deletion hence it has never been reached.
1. Delete only custom license when deleting the dataset; leave the standard ones
2. Fix precommit linting issue


backend/app/routers/datasets.py
```diff
@@ -452,8 +407,14 @@ async def delete_dataset(
         await AuthorizationDB.find(
             AuthorizationDB.dataset_id == PydanticObjectId(dataset_id)
         ).delete()
+
+        # don't delete standard license
+        standard_license_ids = [license.id for license in standard_licenses]
```
ddey2, line 412: Thank you for catching the bug.

I wonder if we really need this check. the license ids for standard licenses are not in the mongodb. Those will never get deleted. But I guess it never hurts to add a check and never call the delete_license function.

Thanks you for runnning the pre-commit properly on my PR changes.

ddey2 (approved):

backend/app/routers/datasets.py
longshuicy: I run the pre-commit run all and fixed more formatting issues. Please check.

ddey2 (approved):
looks good. Don't see any more changes on my local after running pre-commit.

PR #1022, opened 2024-04-24
Enable/disable listeners

ddey2:
Addressed:
1. Adding a page for listing all the extractors
2. Only admins should be able to enable/disable them.
3. Regular users should see the enabled extractors only.

To test:
1. run the test_extractors
2. Try to switch to admin_mode and then enable/disable the extractors
3. Then try to switch back to regular mode and you should see only enabled extractors.
4. Try to disable an extractor, then try to submit a job being a regular user from backend(Postman), you should see Forbidden error.

Adding some screenshots for reference.

Screenshot 2024-05-02 at 5 16 16 PM
Screenshot 2024-05-06 at 3 38 38 PM


tcnichol (approved):
Tested and everything works. Marking approved.

Vismayak:
Tested and working, approved!

Vismayak (approved):

backend/app/routers/listeners.py
```diff
@@ -374,10 +391,66 @@ async def edit_listener(
...
+    return await set_active_flag(listener_id, True)
+
+
+@router.put("/{listener_id}/disable", response_model=EventListenerOut)
+async def disable_listener(
+    listener_id: str,
+    user_id=Depends(get_user),
+    allow: bool = Depends(ListenerAuthorization()),
+):
+    """Enable an Event Listener. Only admins can enable listeners.
```
lmarini, line 414: disable

backend/app/routers/listeners.py
```diff
@@ -374,10 +391,66 @@ async def edit_listener(
...
+
+    Arguments:
+        listener_id -- UUID of the listener to be enabled
```
lmarini, line 417: diabled

backend/app/routers/listeners.py
```diff
@@ -374,10 +391,66 @@ async def edit_listener(
...
+    """
+    return await set_active_flag(listener_id, False)
+
+
+@router.put("/{listener_id}/toggle", response_model=EventListenerOut)
```
lmarini, line 422: if this toggles, do we need the other two methods above?

lmarini (changes requested):
Besides some minor comments there is a couple of high level issues I noticed:

1. Anyone can click on the extractor link, but if you are not in admin mode, nothing shows in the list. Maybe we can have a message for now. Or not show the link in the menu unless you are in admin. I am not sure.
2. Once I went into admin, I clicked on two extractors, but when I go to the dataset submission page, three show active. I countinued to click and the lists on dataset page and file page of enabled extractors are off from the main page.

Thanks!

backend/app/models/listeners.py
```diff
@@ -69,6 +69,7 @@ class EventListenerDB(Document, EventListenerBase):
     modified: datetime = Field(default_factory=datetime.now)
     lastAlive: datetime = None
     alive: Optional[bool] = None  # made up field to indicate if extractor is alive
+    active: Optional[bool] = None
```
lmarini, line 72: This should probably be just `bool` and not optional. It's either enabled or not. Default should be False.

(Same could be said about `alive` above)

lmarini (changes requested):

backend/app/deps/authorization_deps.py
```diff
@@ -389,6 +390,43 @@ async def __call__(
...
+                listener.creator and listener.creator.email == current_user
+            ):
+                return True
+            else:
+                raise HTTPException(
+                    status_code=403,
+                    detail=f"User `{current_user} does not have permission on listener `{listener_id}`",
+                )
+        raise HTTPException(
+            status_code=404, detail=f":Listener {listener_id} not found"
```
lmarini, line 426: Can we remote the starting `:`?

ddey2:
> Besides some minor comments there is a couple of high level issues I noticed:
> 
> 1. Anyone can click on the extractor link, but if you are not in admin mode, nothing shows in the list. Maybe we can have a message for now. Or not show the link in the menu unless you are in admin. I am not sure.
> 2. Once I went into admin, I clicked on two extractors, but when I go to the dataset submission page, three show active. I countinued to click and the lists on dataset page and file page of enabled extractors are off from the main page.
> 
> Thanks!

Hi Luigi,

1. I agree. I was thinking of hiding the link in the menu unless you are an admin and not in admin_mode.
 2. I am not able to reproduce that. I am trying with 2 extractors. It reflects only the enabled extractors on my file/dataset page.

backend/app/deps/authorization_deps.py
```diff
@@ -389,6 +390,43 @@ async def __call__(
         raise HTTPException(status_code=404, detail=f"Group {group_id} not found")
 
 
+class ListenerAuthorization:
+    """We use class dependency so that we can provide the `permission` parameter to the dependency.
+    For more info see https://fastapi.tiangolo.com/advanced/advanced-dependencies/.
+    We want not admins users not to access listeners which are not active"""
```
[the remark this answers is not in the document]
ddey2: done

backend/app/routers/listeners.py
ddey2: I am using this from frontend to toggle the flag. I was thinking if there is a use-case where somebody explicitly calls a backend endpoint to enable/disable the listener irrespective of the active flag. Hence, I created the separate endpoints, but all use the same function internally.

longshuicy: image

After I select the extractors, what do I do to enable/disable? I don't see any button around. I should already be in admin mode.

backend/app/routers/listeners.py
```diff
@@ -374,10 +391,66 @@ async def edit_listener(
...
+@router.put("/{listener_id}/disable", response_model=EventListenerOut)
+async def disable_listener(
+    listener_id: str,
+    user_id=Depends(get_user),
+    allow: bool = Depends(ListenerAuthorization()),
+):
+    """Disable an Event Listener. Only admins can enable listeners.
+
+    Arguments:
+        listener_id -- UUID of the listener to be enabled
```
longshuicy, line 417: These two endpoints seems a little bit redundant? Could it be just one endpoint accepting the parameter then decide to activate or de-activate

backend/app/routers/listeners.py
```diff
@@ -374,10 +391,66 @@ async def edit_listener(
...
+):
+    """Toggle the active flag of an Event Listener. Only admins can enable/disbale listeners.
+
+    Arguments:
+        listener_id -- UUID of the listener to be enabled
+    """
+    listener = await EventListenerDB.find_one(
+        EventListenerDB.id == ObjectId(listener_id)
+    )
+    if listener:
```
longshuicy, line 436: in the dependency class `ListenerAuthorization`, You might be already checking the id exist?

backend/app/routers/listeners.py
ddey2: @longshuicy Clicking on the checkboxes does the job. In this case, both the extractors are enabled.

backend/app/routers/listeners.py
ddey2: Please see my reply to Luigi above. 

If you all think this is redundant, I will get rid of these two.

backend/app/routers/listeners.py
ddey2: I don't if admin and admin_mode both are true

ddey2:
May 16th meeting -
1.  try toggle button instead of checkbox 
2. List all the extractors, disable toggle button for nonadmin mode
3. make set_active_flag method as private and let other endpoints call them

ddey2:
Major update:
1. Use toggle switch button instead of checkbox
2. Showing tooltip for disabled toggle button

For latest UI look, please refer to ss

Screenshot 2024-05-16 at 2 05 16 PM
Screenshot 2024-05-16 at 12 23 04 PM

lmarini (approved):
Looks great! Thanks for making those changes.

PR #1030, opened 2024-04-26
removing alignment css

ddey2:


tcnichol (approved):
Buttons now line up in Firefox, Chrome. Marking approved.

PR #1038, opened 2024-05-06
add generic superadmin flag to endpoints

max-zilla:
This adds a new API parameter called `force_admin` that enables you to make a call as if admin_mode were enabled even if it isn't (assuming you are admin).

Currently on one endpoint:
`http://localhost:8000/api/v2/groups?force_admin=True`

If this is acceptable, we can include anywhere we use the `get_admin_mode` dependency above it:
```
    force_admin: bool = False,
    admin_mode: bool = Depends(get_admin_mode),
```
example endpoint:

http://localhost:8000/api/v2/groups?force_admin=true

will return a list of group data.


backend/app/routers/groups.py
```diff
@@ -34,6 +35,7 @@ async def get_groups(
     user_id=Depends(get_user),
     skip: int = 0,
     limit: int = 10,
+    superadmin: bool = False,
```
longshuicy, line 38: Is this flag being used in the logic?

backend/app/routers/authentication.py
```diff
@@ -116,13 +117,20 @@ async def get_admin(
...
+async def get_admin_mode(
+    superadmin: bool = False, current_username=Depends(get_current_user)
+) -> bool:
     """Get Admin mode from User Object."""
     if (
         current_user := await UserDB.find_one(UserDB.email == current_username.email)
     ) is not None:
-        if current_user.admin_mode is not None:
-            return current_user.admin_mode
+        if current_user.admin:
```
longshuicy, line 127: I'm a little bit confused about the use case. I thought admin_mode is the equivalent of the flag "superadmin"? Would it be possible you share an example of why it is needed here?
Thanks!

max-zilla: we want to be able to pass in a flag to endpoints "superadmin=true" that grants you admin permissions just for that call. we dont want to micromanage turning admin_mode on/off for every call, but the dev team didn't like the idea of leaving it turned on for a user all the time either.

backend/app/routers/groups.py
max-zilla: yes, it is used in `get_admin_mode` function above. this is a param we can add to any function that has `Depends(get_admin_mode)`

backend/app/routers/authentication.py
longshuicy: Ah I understand now. Thank you for explanation!!

max-zilla:
@bingzhang please add `superadmin: bool = False,` to all endpoints in the project that have `admin_mode: bool = Depends(get_admin_mode),` argument.

ddey2:
the logic looks good to me. I see that the PR is in draft state. So I will wait for now.

backend/app/routers/authentication.py
```diff
@@ -116,13 +117,20 @@ async def get_admin(
 
 
 @router.get("/users/me/admin_mode")
-async def get_admin_mode(current_username=Depends(get_current_user)) -> bool:
+async def get_admin_mode(
```
lmarini, line 120: 1. Should we split the dependency from the route and put it in dependencies.py?
2. Should we rename `superadmin` to `admin` or `force_admin` or something else?

max-zilla: I don't quite understand question 1. can you say more about "split the dependency from the route"?

bingzhang: We need to rename `superadmin` to `force_admin`, since `admin` has been used in `admin=Depends(get_admin)`

lmarini: I don't think this was introduced in this PR, but fastapi dependencies don't need to be routes. More importantly this route returns just a boolean and not a JSON document, so we could separate the two and make the route return a valid JSON document. I can make the change.

bingzhang:
> @max-zilla @bingzhang can I rename `force_admin` to `enable_admin`? what do you think?

I am fine with `enable_admin`.

max-zilla:
Resolved merge conflicts and changed to `enable_admin`, a few test API endpoints worked fine.

bingzhang:
@max-zilla I fixed the codegen error.

longshuicy (changes requested):
The logic looks good and I tested the list all dataset endpoint, which works as expected. 

When I do a project-wide search of the admin_mode, I noticed there are one spot missing enable_admin. Could you double check?

```
class FeedAuthorization:
    """We use class dependency so that we can provide the `permission` parameter to the dependency.
    For more info see https://fastapi.tiangolo.com/advanced/advanced-dependencies/.
    Regular users can only see their own feeds"""

    # def __init__(self, optional_arg: str = None):
    #         self.optional_arg = optional_arg

    async def __call__(
        self,
        feed_id: str,
        current_user: str = Depends(get_current_username),
        admin_mode: bool = Depends(get_admin_mode),
        admin: bool = Depends(get_admin),
```

lmarini (approved):

longshuicy (approved):

PR #1046, opened 2024-05-09
List  license and its url on public dataset

longshuicy:
image


backend/app/main.py
```diff
@@ -237,6 +238,9 @@
     tags=["licenses"],
     dependencies=[Depends(get_current_username)],
 )
+api_router.include_router(
```
lmarini, line 241: I wonder if there is a better patterns for public routes? Do we always have to duplicate them?

lmarini (approved):
Left a broad question about public routes. Otherwise seem to work fine.

backend/app/main.py
longshuicy: You can see my latest commit. I don't know how exactly but it works:
https://github.com/clowder-framework/clowder2/pull/1046/commits/1d83fb2b61924f605565e98c4bd0b4e77e938020

PR #1054, opened 2024-05-15
Fixing logic to post metadata on file upload

ddey2:
The root problem was - page was reloaded before it could make call to post metadata. 
Also deleted 2 tsx files which are not used anymore

Steps to reproduce:
1. Create a required metadata field and require it on files
2. Upload two or more  files to a datasets and specify the required metadata
3. Check if the files have that metadata assigned to them


ddey2:
Going to call the route instead of location.reload

longshuicy (approved):
Code looks good. Works well.

PR #1058, opened 2024-05-20
978 migrate users

tcnichol:
This is a draft pull request for now. Here we will migrate users from a clowder v1 instance into a clowder v2 instance. Currently this works for users, datasets, and files inside datasets. 

Next I will be working on folders and metadata. 

A later pull request will address collection hierarchies, CILogon accounts, and other features outside of the scope of this one.

To test this, you will want to run a clowder v1 instance. Since both need to run a different version of MongoDB, change this in the .yml file for the dependencies in v1.

```
  # database to hold metadata (required)
  mongo:
    image: mongo:3.6
    restart: unless-stopped
    networks:
      - clowder
    command: mongod --port 27018
    ports:
      - '27018:27018'
    volumes:
      - mongo:/data/db
```

also change this in application.conf

`mongodbURI = "mongodb://127.0.0.1:27018/clowder"`

If you add entries to the v1 instance, try migrating them.

### DIFF
--- a/backend/Pipfile
+++ b/backend/Pipfile
@@ -33,7 +33,6 @@ pytest = "7.3.1"
 pytest-asyncio = "0.21.0"
 black = "23.3.0"
 faker = "18.4.0"
-python-dotenv = "1.0.1"
 
 [requires]
 python_version = "3.9"
--- a/backend/Pipfile
+++ b/backend/Pipfile
@@ -33,6 +33,7 @@ pytest = "7.3.1"
 pytest-asyncio = "0.21.0"
 black = "23.3.0"
 faker = "18.4.0"
+python-dotenv = "1.0.1"
 
 [requires]
 python_version = "3.9"
--- a/backend/app/keycloak_auth.py
+++ b/backend/app/keycloak_auth.py
@@ -312,7 +312,7 @@ async def get_current_user_id(identity: Json = Depends(get_token)) -> str:
     return keycloak_id
 
 
-async def create_user(email: str, password: str, firstName: str, lastName: str):
+async def create_user(email: str, password: str, firstName: str, lastName: str, temporary=False):
     """Create a user in Keycloak."""
     keycloak_admin = KeycloakAdmin(
         server_url=settings.auth_server_url,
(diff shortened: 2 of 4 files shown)


scripts/migration/migrate.py
```diff
@@ -0,0 +1,372 @@
...
+                file, all_dataset_folders, dataset_v2_id, user_headers_v2
+            )
+    return [USER_MAP, DATASET_MAP]
+
+
+if __name__ == "__main__":
+    # users_v1 = get_clowder_v1_users()
+    USER_MAP = {}
+    DATASET_MAP = {}
+    users_v1 = [
```
longshuicy, line 325: For test. Can comment out

scripts/migration/migrate.py
```diff
@@ -0,0 +1,372 @@
...
+            "firstName": "Chen",
+            "lastName": "Wang",
+            "fullName": "Chen Wang",
+            "email": "cwang138-clowder2@illinois.edu",
+            "avatar": "http://www.gravatar.com/avatar/2f97a52f2214949c4172d7fb796f173e?d=404",
+            "profile": {},
+            "identityProvider": "Chen Wang (cwang138@illinois.edu) [Local Account]",
+        }
+    ]
+    users_v1 = get_clowder_v1_users()
```
longshuicy, line 343: For test all the users

tcnichol:
On spaces, right now if you want to test spaces you'll need to do it locally with this branch:

https://github.com/clowder-framework/clowder/pull/453

Also add the changes to v1 mentioned above. 

Currently v1 api does not return creator of spaces with space api calls. The above pull request adds that.

scripts/migration/migrate.py
```diff
@@ -0,0 +1,377 @@
...
+    # TODO what about pagination
+    endpoint = f"{CLOWDER_V1}/api/datasets?limit=0"
+    response = requests.get(endpoint, headers=clowder_headers_v1, verify=False)
+    return [dataset for dataset in response.json() if dataset["authorId"] == user_id]
+
+
+def get_clowder_v1_user_spaces(user_v1):
+    endpoint = f"{CLOWDER_V1}/api/spaces"
+    response = requests.get(endpoint, headers=clowder_headers_v1, verify=False)
+    return [space for space in response.json() if space["creator"] == user_v1["id"]]
```
ddey2, line 82: Does it have creator?

ddey2: Ignore the above comment. I found your other PR adding the creator in Clowder1.

PR #1060, opened 2024-05-21
1053 pagination doesnt update on deleting files in a dataset

longshuicy:
Now delete will trigger a get request to the backend to update the new list. 
- I created new redux state (deletedXXXX)
- whenever item got deleted this state will change
- Upon listening to such change, a request to the backend to fetch the latest list of resources
- I also fixed other deletes in the same fashion for API keys, groups, dataset, and metadata definitions.

---

To test:
1. Create many files/folders, and delete. The items as well as the pagination should reflect. 
2. Delete datasets
3. Delete groups. I added a delete button right next to each group
4. Delete API keys
5. Delete metadata definition

### DIFF
--- a/frontend/src/components/Explore.tsx
+++ b/frontend/src/components/Explore.tsx
@@ -31,6 +31,9 @@ export const Explore = (): JSX.Element => {
 	const datasets = useSelector(
 		(state: RootState) => state.dataset.datasets.data
 	);
+	const deletedDataset = useSelector(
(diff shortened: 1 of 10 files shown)


lmarini:
I think I understand the issue and how you fixed, but I am on main and I can't reproduce :(

If I delete a dataset or a group the list gets updated. In the case of dataset even if it's not on the first page.

lmarini (approved):

PR #1063, opened 2024-05-28
1040 delete given dataset version

longshuicy:
This PR:

1. Rewrite delete file, delete dataset, and delete released dataset logic
2. Wrap common operations into utility functions and reuse
3. Frontend delete interface

----

The protocol of **deleting a released dataset** is listed as below

1. Delete associate metadata (mongo doc)
2. Delete folders (mongo)
3. Delete dataset thumbnail (mongo &  raw bytes if applicable)
4. Delete dataset visualization config (mongo)
5. Delete dataset visualization data (mongo & raw bytes if applicable)
6. Delete files
    6.1. Delete file metadata (mongo)
    6.2. Delete file thumbnail (mongo & raw bytes if applicable)
    6.3. Delete file vis config (mongo)
    6.4. Delete file vis data (mongo & raw bytes if applicable)
    6.5. Delete file (mongo & raw bytes if applicable)
7. Mark released dataset deleted (mongo)

----

To **delete a current dataset**
1. find all the **associate released dataset** and delete those following above steps
2. delete current metadata, folder, thumbnail, visualization and files.

image


lmarini (changes requested):
One potential issue:

- I have created a dataset, created a version, deleted a file. If I go back to the versioned dataset  and click on any of the files I get access denied. Could you try to reproduce? Thank you.

longshuicy:
> One potential issue:
> 
> * I have created a dataset, created a version, deleted a file. If I go back to the versioned dataset  and click on any of the files I get access denied. Could you try to reproduce? Thank you.

I can't reproduce this issue. Maybe we can look at it together at some time. 
Below are what I tried:

- upload files and create a version
- delete the file and go back to a version --> file still accessible

---
- upload files and create version
- delete the version --> file still exist in the latest 

--- 
- upload files and create several versions
- delete selected version --> file still exist in other versions

---
- hard refresh and files still all accessible

ddey2:
@lmarini  I couldn't reproduce your issue.

backend/app/db/dataset/version.py
```diff
@@ -165,3 +173,298 @@ async def _freeze_dataset_visualization(
...
+            is None
+        ) and (
+            await FileDB.find_one(FileDB.id == PydanticObjectId(frozen_file.origin_id))
+        ) is None:
+            fs.remove_object(settings.MINIO_BUCKET_NAME, str(frozen_file.origin_id))
+
+    return frozen_file.dict()
+
+
+async def _delete_frozen_dataset(frozen_dataset, fs, hard_delete=False):
```
ddey2, line 326: when is hard_delete set to True?

backend/app/db/dataset/version.py
```diff
@@ -165,3 +173,298 @@ async def _freeze_dataset_visualization(
...
+    return frozen_dataset.dict()
+
+
+# TODO: Move this to MongoDB middle layer
+async def remove_file_entry(
+    file_id: Union[str, ObjectId],
+    fs: Minio,
+    es: Elasticsearch,
+):
+    """Remove FileDB object into MongoDB, Minio, and associated metadata and version information."""
```
ddey2, line 374: Can you add the parameters and return type to the doc string? May be add them to all the public functions?

backend/app/db/dataset/version.py
```diff
@@ -165,3 +173,298 @@ async def _freeze_dataset_visualization(
...
+
+        # delete file raw bytes if not used anywhere else
+        await _delete_file(file, fs)
+
+    else:
+        raise HTTPException(status_code=404, detail=f"File {file_id} not found")
+
+
+async def remove_frozen_file_entry(frozne_file_id: Union[str, ObjectId], fs: Minio):
+    """Remove Frozen FileDB object into MongoDB, Minio, and associated metadata and version information."""
```
ddey2, line 403: Are you deleting the file from the Minio here in this function?

backend/app/db/dataset/version.py
```diff
@@ -165,3 +173,298 @@ async def _freeze_dataset_visualization(
...
+
+        # if all above succeeded
+        await _delete_file(file, None)
+
+    else:
+        raise HTTPException(status_code=404, detail=f"File {file_id} not found")
+
+
+async def remove_frozen_local_file_entry(frozen_file_id: Union[str, ObjectId]):
+    """Remove FileDB object into MongoDB, Minio, and associated metadata and version information."""
```
ddey2, line 452: Are you deleting the file from the Minio here in this function?

ddey2: All these functions look very similar. Can you add a little more doc?
I am not able to understand the requirement for 4 separate functions apart from the fact that 2 of them are for frozen files.

frontend/src/components/datasets/versions/DatasetVersions.tsx
```diff
@@ -0,0 +1,243 @@
...
+		history(`/datasets/${currDataset.origin_id}?currVersionPageNum=1`);
+		setDeleteDatasetVersionConfirmOpen(false);
+	};
+
+	return (
+		<>
+			<ActionModal
+				actionOpen={deleteDatasetVersionConfirmOpen}
+				actionTitle="Delete Dataset"
+				actionText="Are you sure you want to delete this version? This process cannot be undone.
```
ddey2, line 117: I was wondering if we can have this text in some kind of config file. So that it's easier to change the text in a single place. No need to address this here. We can think about this later.

ddey2 (changes requested):
1. I tried creating multiple versions. There seem to be some issue with the timestamp with creation datetime. All are exactly same though I created them few minutes apart. 
Screenshot 2024-06-07 at 11 49 54 AM
2. Also, why do we still want to keep record of deleted dataset version? I see we delete teh associated files and folders but keep the dataset version and show them as deleted.
3. Can you please add testcases? This is a big change.

backend/app/db/dataset/version.py
longshuicy: When deleting the whole dataset (Latest), it will be set to true to ensure all the bytes in minio are gone as well. 
In the code, the place it is set to true is on the Delete endpoint:

```
@router.delete("/{dataset_id}")
async def delete_dataset(
...
 # find associate frozen datasets and delete them iteratively
        async for frozen_dataset in DatasetFreezeDB.find(
            DatasetFreezeDB.origin_id == PydanticObjectId(dataset_id)
        ):
            await _delete_frozen_dataset(frozen_dataset, fs, hard_delete=True)
```

backend/app/db/dataset/version.py
longshuicy: Yes, I'm deleting the raw bytes in this private function 
```  
# delete file raw bytes if not used anywhere else
await _delete_file(file, fs)
```

backend/app/db/dataset/version.py
longshuicy: Yes same as above, I'm deleting them in here
```
await _delete_frozen_file(frozen_file, fs)
```

backend/app/db/dataset/version.py
longshuicy: Yes I will add more descriptions. 
This is for local files which does not have minio backup. I didn't implement that feature so I am not 100% sure what's the usecase. 
But to keep it consistent with the original implementation,  I'm not passing in the `fs` hence not deleting any raw bytes.
```
await _delete_frozen_file(frozen_file, None)
```

frontend/src/components/datasets/versions/DatasetVersions.tsx
longshuicy: Created an issue for that: https://github.com/clowder-framework/clowder2/issues/1100

backend/app/db/dataset/version.py
longshuicy: I added doc strings to all the functions here.

backend/app/db/dataset/version.py
longshuicy: doc strings added.

longshuicy:
> 1. tried creating multiple versions. There seem to be some issue with the timestamp with creation datetime. All are exactly same though I created them few minutes apart.

1. There isn't any timestamp for the release; I was using the last modified timestamp on the dataset.
If there is no modification, there won't be any change. I'm adding a "last modified" to make it clearer. 
2. We keep the empty record just to keep a linear version history, and to indicate there has deletion happened.
3. Added a new pytest to test dataset versioning specifically

image

ddey2 (changes requested):
Thank you so much for adding the test cases. It will help things from breaking in future and really helps to review too. :)
Could you add a few test cases?
1. Could you add test for visualization and vis_config like you did for thumbnails and metadata?

2. Once you delete the whole dataset, can you make sure all the original bytes are all deleted?

longshuicy:
> Thank you so much for adding the test cases. It will help things from breaking in future and really helps to review too. :) Could you add a few test cases?
> 
> 1. Could you add test for visualization and vis_config like you did for thumbnails and metadata?
> 2. Once you delete the whole dataset, can you make sure all the original bytes are all deleted?

Added

backend/app/tests/test_dataset_versioning.py
```diff
@@ -127,6 +173,31 @@ def test_freeze_and_delete(client: TestClient, headers: dict):
...
+    assert response.status_code == 404
+    response = client.get(
+        f"{settings.API_V2_STR}/visualizations/{version_1_vis_id}/",
+        headers=headers,
+    )
+    assert response.status_code == 404
+
     # delete the whole dataset
     response = client.delete(
         f"{settings.API_V2_STR}/datasets/{dataset_id}", headers=headers
```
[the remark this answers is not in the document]
longshuicy: Thank you for the suggestion!

lmarini (approved):

PR #1064, opened 2024-05-30
1061 filesystem UI component

Vismayak:
To test:
Run an extractor, preferably locally to do the necessary modifications:
  1. Add a parameter in `extractor-info.json` file,  in the `schema` field of  `parameters` 
  ```json
      "DATASET_FILE_ID":{
              "type":"string",
              "title":"Dataset file ID",
              "description":"The Clowder file id of the dataset file to be used for inference",
              "format": "clowderFile"
         },
  ```
  2. The extractor should have a field like this
  image


frontend/src/components/navigation/FileSelector.tsx
```diff
@@ -0,0 +1,288 @@
...
+	const [FSItems, setFSItems] = useState<FSItem[]>([]);
+
+	// API function call to get Datasets
+	const listDatasets = (skip?: number, limit?: number, mine?: boolean) => {
+		dispatch(fetchDatasets(skip, limit, mine));
+	};
+
+	// Fetch datasets on component mount
+	useEffect(() => {
+		listDatasets(0, 10, true);
```
longshuicy, line 190: You are fixing the dataset to be 10? Could you add a todo here to address dataset pagination.

longshuicy (commented):
Could you add some example how do I trigger this widget? Do I need a register a special extractor with specific fields? Thanks!

Vismayak:
@longshuicy sorry for the confusion, I have updated the testing instructions, let me know if the updated instructions are better!

frontend/src/components/navigation/FileSelector.tsx
```diff
@@ -0,0 +1,289 @@
...
+}) => {
+	const [expanded, setExpanded] = useState(false);
+	const [children, setChildren] = useState<FSItem[] | undefined>(item.children);
+	const isFolder = item.type === "folder";
+
+	// Function to generate Icon based on item type
+	const getIcon = () => {
+		if (isFolder) {
+			return <FolderIcon />;
+		}
```
longshuicy, line 62: If possible can we differentiate it with actual folder?
For dataset, we use this icon `import DatasetIcon from "@mui/icons-material/Dataset";`. We use this in the search page.

e.g. "test new dataset delete" is a dataset, and "folder renamed" is actually a folder.
image

longshuicy (commented):
Thanks for the write-ups on how to test. 
Very cool it is able to pick up the file id and send it to the extract!
I have just a few comments. Let me know if it's not very straightforward to address; then maybe you can create another issue in that case.

ddey2 (changes requested):
It looks very nice. I agree with @longshuicy I think we should show diff icons for datasets and folders to avoid confusion. Rest looks all good.

longshuicy (approved):
Works well.

ddey2 (approved):
dataset and folder icons look consistent with what we have

PR #1090, opened 2024-06-06
1066 index folders

longshuicy:
1. Fix bugs - when dataset/files updates/downloads, need to reindex. (Now the filters by download actually works, previously it was not)
2. Add folder to index; use the dataset status for public search
3. Frontend 
  - sort in descend order by created date
  - add folder to search results
  
image
image


ddey2:
@longshuicy Could you take a look at the pytest failure?

lmarini (approved):

PR #1103, opened 2024-06-10
1086 file UI component flag current files in datasets

Vismayak:
## Aim 
To create an option in the file component viewer to only see the files in the current dataset

## Methodology

The difficulty behind the task is because the datasetID depends on which dataset the extractor is called from and therefore can not be hardcoded into the `extractor_info` json. In order to bypass, this we use the property [`uiSchema`](https://rjsf-team.github.io/react-jsonschema-form/docs/api-reference/uiSchema/) (that I have previously misunderstood) and `"ui:options"`.

 uiSchema is actually used to format a given input and is dependent on the parameter name. `ui:options` allows us to pass in the datasetID as a field. Since the uiSchema uses parameter names I use a for loop to iterate through the form's parameters and add to the uiSchema

```js
	// The for loop is used to pass the datasetId to a widget if it is a clowderFile Widget
	if (
		selectedExtractor &&
		selectedExtractor["properties"] &&
		selectedExtractor["properties"]["parameters"] &&
		selectedExtractor["properties"]["parameters"]["schema"]
	) {
		const parameters = selectedExtractor["properties"]["parameters"]["schema"];
		for (const key in parameters) {
			// Check if field "format" present and equals "clowderFile"
			if (parameters[key].format && parameters[key].format === "clowderFile") {
				uiSchema[key] = {
					"ui:widget": "clowderFile",
					"ui:options": {
						datasetId: datasetId,
					},
				};
			}
		}
	}
```

The rest of the changes are pretty straightforward. It involves passing in the datasetID and a flag to only show the dataset files or the entire filesystem.

## How to test

Run an extractor, I used [wordcount](https://github.com/clowder-framework/clowder2-demos/tree/main/word-cloud-extractor) , locally as in to do the necessary modifications:

  1. Add another parameter in `extractor-info.json` file,  in the `schema` field of  `parameters`, if using wordcount, you will have to add it [here](https://github.com/clowder-framework/clowder2-demos/blob/0ba84d236d98290db5b9e75f6749fed52eee857d/word-cloud-extractor/extractor_info.json#L38) 
  ```js
  "DATASET_FILE_ID": {
          "type": "string",
          "title": "Dataset file ID",
          "description": "The Clowder file id of the dataset file to be used for inference",
          "format": "clowderFile",
          "showOnlyDatasetFiles": true
        }
  ```

3. When you open the extractor and open the fileselector, you should only see the current folders and files in the dataset
  image

4. Re-run the extractor but tis time set the field "showOnlyDatasetFiles" to false or remove it
5. You should then see the complete clowder filesystem
image


frontend/src/components/styledComponents/ClowderFileSelector.jsx
```diff
@@ -6,10 +6,17 @@ export const ClowderFileSelector = (item) => {
 	const handleChange = (value) => {
 		item.onChange(value);
 	};
+	const datasetId = item.options.datasetId;
+	const showOnlyDatasetFiles = item.schema.showOnlyDatasetFiles ? true : false;
+	console.log(showOnlyDatasetFiles);
```
longshuicy, line 11: Could you take out the console.log

longshuicy (commented):
except minor suggestion everything else works well.

frontend/src/components/navigation/FileSelector.tsx
```diff
@@ -47,6 +47,41 @@ interface FSItem {
 	content_type?: string;
 }
```
longshuicy, line 49: Actually on a second thought; on this widget, can you read the URL query parameter to get the current datasetid instead? This way you can avoid changing the submitExtraction.tsx
There is a pattern usually: 
http://localhost:3000/files/66671d1e15f93923fbe8b110?dataset=6660b59f23f92e24d2f8bb40&folder=null&verNum=6669fdff1d5e4ec1df0257cf

Your "best friend" suggested:
```
function getDatasetId(url) {
    const urlObj = new URL(url);
    const datasetId = urlObj.searchParams.get("dataset");
    return datasetId;
}
```

longshuicy (approved):

PR #1104, opened 2024-06-10
1098 superamin messes up dataset pagination

tcnichol:
A fix that has worked for me is to make sure that mine=false when using superadmin mode.


lmarini (changes requested):
- I think this fixes the listed datasets, but the pagination controls seem to have issues. The number of pages is updated, but I don't think a page is selected
- What are the changes to `open.json` for?

tcnichol:
> * I think this fixes the listed datasets, but the pagination controls seem to have issues. The number of pages is updated, but I don't think a page is selected
> * What are the changes to `open.json` for?

I think open.json got changed when I ran the pre-commit hooks. I can undo that. 

Will check page pagination.

longshuicy (changes requested):
Admin mode pagination seems to be fixed.
However, in admin mode when I try to switch to "my dataset" tab, it doesn't allow me to flip pages. When I check the network tab I see double calls to the backend: one with mine=true, the other with mine=false.

image

This might related to the logic you added, when adminMode = True you list the dataset with mine=false

```
if (adminMode) {
			listDatasets(0, limit, false);
```
This part.

ddey2 (changes requested):
Pagination doesn't work on 'My dataset' tab in superadmin mode. 
Screenshot 2024-06-24 at 1 45 46 PM

tcnichol:
> Pagination doesn't work on 'My dataset' tab in superadmin mode. Screenshot 2024-06-24 at 1 45 46 PM

I think I fixed this.

I was not using the current page num in listDatasets, so it was getting stuck at 0 for admin mode and 'mine.' Now when you switch tabs, it changes 'mine' and 'currentPageNum' and then those are used in the hook to populate the datasets.

longshuicy:
@tcnichol whenever you have a chance could you check if this is resolved with your latest change in the PR? Thanks!! https://github.com/clowder-framework/clowder2/issues/1123

frontend/src/components/Explore.tsx
```diff
@@ -61,11 +66,11 @@ export const Explore = (): JSX.Element => {
 		if (newTabIndex === 1) {
 			setMine(true);
 			setCurrPageNum(1);
-			listDatasets(0, limit, true);
+			// listDatasets(0, limit, true);
```
longshuicy, line 69: could you remove the inline comment. same as below?

longshuicy (approved):
This is working correctly now. Also it resovled this issue https://github.com/clowder-framework/clowder2/issues/1123

ddey2 (approved):
Works perfect.

PR #1111, opened 2024-06-14
1082 group is not updated after delete user

tcnichol:
This is draft, I think I found a fix but my pre commit is still acting up.

### DIFF
--- a/frontend/src/components/groups/Group.tsx
+++ b/frontend/src/components/groups/Group.tsx
@@ -34,6 +34,9 @@ export function Group() {
 		dispatch(deleteGroup(groupId));
 
 	const groupAbout = useSelector((state: RootState) => state.group.about);
+	const groupOwner = useSelector(
+		(state: RootState) => state.group.about.creator
+	);
 	const role = useSelector((state: RootState) => state.group.role);
 
 	const [addMemberModalOpen, setAddMemberModalOpen] = useState(false);
@@ -93,6 +96,7 @@ export function Group() {
 				}}
 				groupName={groupAbout.name}
 				groupId={groupAbout.id}
+				groupOwner={groupOwner}
 			/>
 			<Box
 				sx={{
--- a/frontend/src/components/groups/Group.tsx
+++ b/frontend/src/components/groups/Group.tsx
@@ -1,44 +1,39 @@
 import React, { useEffect, useState } from "react";
-import { Box, Button, Grid, Link } from "@mui/material";
+import { Box, Button } from "@mui/material";
 import Layout from "../Layout";
 import { RootState } from "../../types/data";
 import { useDispatch, useSelector } from "react-redux";
-import { fetchGroupAbout } from "../../actions/group";
+import { deleteGroup, fetchGroupAbout } from "../../actions/group";
 import { fetchGroupRole } from "../../actions/authorization";
 import Typography from "@mui/material/Typography";
-import { useParams } from "react-router-dom";
+import { useNavigate, useParams } from "react-router-dom";
 import { AuthWrapper } from "../auth/AuthWrapper";
 import PersonAddAlt1Icon from "@mui/icons-material/PersonAddAlt1";
 import MembersTable from "./MembersTable";
-import { EditMenu } from "./EditMenu";
 import AddMemberModal from "./AddMemberModal";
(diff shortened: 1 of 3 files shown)


tcnichol:
I think something is messed up here. I only wanted changes in Group component, I reverted changes in the other one, but it still shows changes.
Draft till I figure this out.

longshuicy:
I switched this to draft according to your comments; if this is ready feel free to switch it back. Thanks!

ddey2:
@tcnichol I see this

Screenshot 2024-06-28 at 12 26 33 PM

ddey2:
@tcnichol You might want to merge main to resolve the above error. 

Also, it doesn't work. I deleted user but it doesn't update it on the page unless you refresh it.

ddey2 (changes requested):
@tcnichol  You might want to merge main to resolve the above error.

Also, it doesn't work. I deleted user but it doesn't update it on the page unless you refresh it.

tcnichol:
> @tcnichol You might want to merge main to resolve the above error.
> 
> Also, it doesn't work. I deleted user but it doesn't update it on the page unless you refresh it.

Can you try to delete the node_modules file and then do npm install again? When I switched to another branch and back, it wasn't working until I did npm install again.

frontend/src/components/groups/Group.tsx
```diff
@@ -1,103 +1,82 @@
...
-		},
-	];
+	}, [groupAbout]);
 
 	return (
 		<Layout>
-			{/*Error Message dialogue*/}
-			<ErrorModal errorOpen={errorOpen} setErrorOpen={setErrorOpen} />
-			{/*breadcrumb*/}
-			<Grid container>
```
longshuicy, line 74: any reason you are removing the breadcrumb?

frontend/src/components/groups/Group.tsx
```diff
@@ -1,103 +1,82 @@
...
+			<ActionModal
+				actionOpen={deleteGroupConfirmOpen}
+				actionTitle="Are you sure?"
+				actionText="Do you really want to delete this group? This process cannot be undone."
+				actionBtnName="Delete"
+				handleActionBtnClick={() => {
+					groupDeleted(groupId);
+					setDeleteGroupConfirmOpen(false);
+					// Go to Explore page
+					history("/");
```
longshuicy, line 83: Is it ok to go to /groups page instead?

frontend/src/components/groups/Group.tsx
tcnichol: No, this must have been a mistake. I have put it back in.

frontend/src/components/groups/Group.tsx
```diff
@@ -149,16 +143,25 @@ export function Group() {
 								setAddMemberModalOpen(true);
 							}}
 							endIcon={<PersonAddAlt1Icon />}
-							sx={{ marginRight: "0.5em", width: "auto" }}
 						>
 							Add Member
 						</Button>
 					</AuthWrapper>
-					<AuthWrapper currRole={role} allowedRoles={["owner", "editor"]}>
-						<EditMenu setDeleteGroupConfirmOpen={setDeleteGroupConfirmOpen} />
```
longshuicy, line 158: Any reason you are get rid of the edit menu which includes "edit group names etc"?

longshuicy (changes requested):
This PR works but it seems to get rid of a few existing functionality. 
Also a little explanation on what's the culprit of the issue and what's the fix would be greatly appreciated!

frontend/src/components/groups/Group.tsx
tcnichol: I think this was a mistake somehow. I will fix this and check to see what else is missing from main.

tcnichol:
@longshuicy 

I went back and reverted all of these files to main, and then just made the change necessary to make this work. I'm not sure why it got changed so much.

The basic thing was the useEffect hook needed to depend on groupAbout, so it would populate the group members correctly when groupAbout changed after a user was deleted from the group. Now it updates right for delete and add, and that should be the only change.

longshuicy:
Thanks for the explanation. Deletion works well now.

However when I try to edit the role of any user, it wouldn't allow me to select any of those selection
image

longshuicy (approved):

PR #1112, opened 2024-06-17
1069 show message when no metadata is present

tcnichol:
I added some messages here if there is no extractor or user metadata, for datasets or files. 

@lmarini I used the text you had supplied in the issue for the user metadata, and came up with something else for machine metadata. I think the text in those spaces can be better, but I wanted this to be a draft now so others can weigh in on the messages.


frontend/src/components/metadata/DisplayMetadata.tsx
```diff
@@ -13,6 +13,7 @@ import {
 import { Agent } from "./Agent";
 import { MetadataDeleteButton } from "./widgets/MetadataDeleteButton";
 import { fetchPublicDatasetMetadata } from "../../actions/public_dataset";
+import { Link as RouterLink } from "react-router-dom";
```
longshuicy, line 16: I didn't find where the RouterLink is being used. Could you remove that?

frontend/src/components/metadata/DisplayMetadata.tsx
```diff
@@ -117,61 +118,76 @@ export const DisplayMetadata = (props: MetadataType) => {
...
-						}
+									</Box>
+								);
+							}
+						});
 					});
-				});
+				} else {
+					return (
+						<Box textAlign="center">
```
longshuicy, line 179: The center looks a bit weird. Could it be just align on the left? 
Thoughts? 
image

longshuicy (changes requested):
Coud you resolve merge conflict?

ddey2 (changes requested):
It always says 'no user metadata provided for dataset', even for files.
Screenshot 2024-06-28 at 12 22 39 PM

frontend/src/components/metadata/DisplayMetadata.tsx
tcnichol: Aligned left.

tcnichol:
> It always says 'no user metadata provided for dataset', even for files. Screenshot 2024-06-28 at 12 22 39 PM

I changed 'dataset' to 'resource' so that both messages just say 'resource'.

frontend/src/components/metadata/DisplayMetadata.tsx
tcnichol: removed.

longshuicy (approved):
Looks good.

PR #1113, opened 2024-06-17
Removed snackbar message on clearing metadata form

ddey2:
Try create a metadata definition. In last step, click on 'Clear' the form instead of submit, you should no longer see any snackbar. 

I have also removed the snackbar on successfully creating a definition. Since it takes you to the newly created metadata defn page, the user can't really see the snackbar and I don't see the use of the snackbar there.


longshuicy (approved):
Works as described.

PR #1116, opened 2024-06-20
added check for rendering step1

ddey2:
1. Make sure you have no metadata defn for files
2. Try to upload a file
3. You should not see the step1 saying "Fill in metadata"

Current UI:
Screenshot 2024-06-20 at 11 36 45 AM

### DIFF
--- a/frontend/src/components/files/UploadFileDragAndDrop.tsx
+++ b/frontend/src/components/files/UploadFileDragAndDrop.tsx
@@ -123,6 +123,24 @@ export const UploadFileDragAndDrop: React.FC<UploadFileDragAndDropProps> = (
 		});
 	};
 
+	const checkIfMetadataNeeded = () => {
+		// This function checks if Step 1 to upload metadata is needed when users upload new file
+		let required = false;
+
+		metadataDefinitionList.forEach((val, _) => {
+			if (val.required_for_items.files) {
+				val.fields.forEach((field) => {
+					if (field.required) {
+						required = true;
+						return;
+					}
+				});
+			}
+		});
+
+		return required;
+	};
+
 	// step 1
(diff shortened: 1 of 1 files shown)


frontend/src/components/files/UploadFileDragAndDrop.tsx
```diff
@@ -184,32 +184,37 @@ export const UploadFileDragAndDrop: React.FC<UploadFileDragAndDropProps> = (
...
-								<Grid xs={11}>
-									<Button
-										variant="contained"
-										onClick={handleNext}
-										disabled={!checkIfFieldsAreRequired() ? false : !allFilled}
-										sx={{ display: "block", marginLeft: "auto" }}
-									>
-										Next
-									</Button>
+					{metadataDefinitionList?.length > 0 && (
```
longshuicy, line 187: I don't have any file metadata but I do have some dataset metadata. When uploading files it still prompt me to fill metadata
image

I think merely check metadataDefinitionList might not be enough since it doesn't differentiate dataset metadata vs file metadata; Nor does it differentiate "required" vs "not required".

ddey2: We need to decide if we want to consider those. Right now I just checked if there is no metadata defn, it should not show Step 1. If there is any optional metadata defn, users might still want to show them.

longshuicy (commented):
Could you check now even if metadata is required it's not showing the first step. See screenshot:

image

ddey2:
@longshuicy It will not show up as the required metadata has no required field.

longshuicy:
> @longshuicy It will not show up as the required metadata has no required field.

Ah my bad. I understand now. Approved.

longshuicy (approved):

PR #1117, opened 2024-06-20
Removing version column from File entry view

ddey2:
Current file entry should look like this:
Screenshot 2024-06-20 at 11 48 54 AM


tcnichol (approved):
This looks fine I'm marking approved.

longshuicy (commented):
Could you match folder to match the column number 
image

ddey2:
> Could you match folder to match the column number image

done. Good catch, thanks!

longshuicy (approved):
Works as expected now.

PR #1120, opened 2024-06-21
Download should be incremented as soon as you click on downloads

ddey2:


tcnichol:
@ddey2 

I am testing this for datasets and files.

With datasets, when I click the download button the download increments, but the dataset did not download. This happens both on the regular dataset view and the public dataset view.

frontend/src/components/datasets/ActionsMenu.tsx
```diff
@@ -32,7 +38,7 @@ export const ActionsMenu = (props: ActionsMenuProps): JSX.Element => {
 			<Button
 				sx={{ minWidth: "auto" }}
 				variant="contained"
-				href={`${config.hostname}/api/v2/datasets/${datasetId}/download`}
+				onClick={() => download(datasetId)}
```
longshuicy, line 41: @lmarini I recall we intentionally changed from the download function to just embed href to handle large file download at a point: https://github.com/clowder-framework/clowder2/pull/692
Do we have to change it back?

longshuicy (changes requested):
I don't think download works. I see it in the network with the binaries; but since you are not catch the binary and save it to blob then generate a tag, the binary is not accessible to end user. 

The public download doesn't work for the same reason.

We used to take similar approach as I described before, but at one point we switch to just embed the url and let the browser handle download due to large files. Not sure if we would like to switch back. Maybe something we can discuss?

ddey2:
this is fixed. @tcnichol @longshuicy feel free to take a look

tcnichol:
Tested.

From the dataset page, every click of download increments. Same for files.

If I download datasets from 'explore' download also increments. Same with downloading files from the filelist.

Marking this one approved.

tcnichol (approved):
Download increments on datasets and files from any location. Marking this approved.

longshuicy (commented):
Works well. Thanks!

longshuicy (approved):

PR #1133, opened 2024-07-01
removing left dividers

tcnichol:
The bug was that if you were a read only user, you saw 2 dividers that were empty on the left.

Now you don't see those 2 dividers, but the 'search' section looks larger than it should.

Not sure how to fix it, so this is draft.
left_panel_issues


longshuicy (changes requested):
Could we just remove the divider altogether? I don't think it will look too bad without it. Thoughts?

tcnichol:
@longshuicy i have removed the left dividers and fixed the spacing.

Here is a screenshot of it for logged in (not read only) users
left_tab

tcnichol:
Here is the left drawer for the read only user:
readonly_user_left_tab

ddey2 (approved):

longshuicy (approved):

PR #1136, opened 2024-07-02
Use the update flag when updating fields of existing Elasticsearch entry

longshuicy:
This PR add the correct "update" flag to places where elasticsearch entry already existed. 
Fixed a bug related to "update dataset metadata".
Fixed a few parameters to its correct data type e.g. DatasetOut, FileOut

### DIFF
--- a/backend/app/routers/metadata_datasets.py
+++ b/backend/app/routers/metadata_datasets.py
@@ -229,8 +229,9 @@ async def update_dataset_metadata(
 
         md = await MetadataDB.find_one(*query)
         if md is not None:
+            patched_metadata = await patch_metadata(md, content, es)
             await index_dataset(es, DatasetOut(**dataset.dict()), update=True)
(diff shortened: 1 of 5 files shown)


backend/app/routers/metadata_datasets.py
```diff
@@ -229,8 +229,9 @@ async def update_dataset_metadata(
 
         md = await MetadataDB.find_one(*query)
         if md is not None:
-            await index_dataset(es, dataset, update=True)
-            return await patch_metadata(md, content, es)
+            patched_metadata = await patch_metadata(md, content, es)
+            await index_dataset(es, DatasetOut(**dataset.dict()), update=True)
+            return patched_metadata
```
longshuicy, line 234: There was a bug here. The metadata is updated after indexing hence the updated metadata never reflected

ddey2 (approved):
looks good. Good catch for the bug!

PR #1140, opened 2024-07-05
error message when sharing dataset with user

tcnichol:
Previously when you shared a dataset with a user, it showed an error message but the user was added.

I have fixed this so now you get the success message.


ddey2 (commented):
This looks okay. 

I was wondering why it was previously needed to explicitly check for user role and verify if the user was added on frontend. Looks like @tcnichol you wrote this code. Do you remember?

frontend/src/components/datasets/ShareDatasetModal.tsx
```diff
@@ -85,34 +85,11 @@ export default function ShareDatasetModal(props: ShareDatasetModalProps) {
...
+		setShowSuccessAlert(true);
 		getRoles(datasetId);
-		const currentRoles = datasetRoles;
-		// TODO check for roles
-		let found_user = false;
-		if (currentRoles !== undefined && currentRoles.user_roles !== undefined) {
-			currentRoles.user_roles.map((currentRole) => {
-				if (currentRole.user.email === email) {
-					if (currentRole.role.toString() === role.toString()) {
-						found_user = true;
```
longshuicy, line 99: Is this logic not needed anymore?

tcnichol: It is not.

frontend/src/components/datasets/ShareDatasetModal.tsx
```diff
@@ -85,34 +85,11 @@ export default function ShareDatasetModal(props: ShareDatasetModalProps) {
 	}, [users, myProfile.email]);
 
 	const onShare = async () => {
-		setShowFailAlert(false);
-		setShowSuccessAlert(false);
 		await setUserRole(datasetId, email, role);
+		// setGroup({ label: "", id: "" });
+		setRole("viewer");
+		setShowSuccessAlert(true);
```
longshuicy, line 91: I don't quite understand this part `setRole("viewer");`, why hard code to "viewer"?

tcnichol:
> This looks okay.
> 
> I was wondering why it was previously needed to explicitly check for user role and verify if the user was added on frontend. Looks like @tcnichol you wrote this code. Do you remember?

I think if I didn't do this the page did not show the user that was added correctly.

frontend/src/components/datasets/ShareDatasetModal.tsx
tcnichol: I checked. This is there because if you add a user with another permission, and then you go to add a different user, the default permission will not be 'viewer' but whatever permission you added last. I was thinking that the default any time you go to add someone was supposed to be viewer?

longshuicy: I see i understand now. Thanks!

longshuicy (approved):

PR #1153, opened 2024-07-24
setting active to it's original value for existing extractor

ddey2:


tcnichol (approved):
I tested this out and it works. I'm marking approved.

longshuicy:
I notice below behavior:
- When a dead extractor being enabled (see below) I can see it in the extractors under analysis tab
- When this extractor become alive, it disappeared from the analysis tab as well as becomes disabled automatically
image

lmarini (approved):

PR #1163, opened 2024-07-31
error in creating datasets when using docker-compose, newer backend tag used

tcnichol:
Using docker compose during the hackathon I noticed an error would happen when users create datasets. The reason is that the image used for the backend did not include license feature, so I used a later release here.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -37,7 +37,7 @@ services:
       - /var/run/docker.sock:/var/run/docker.sock
 
   backend:
-    image: 'clowder/clowder2-backend:2.0.0-beta.2'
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -37,7 +37,7 @@ services:
       - /var/run/docker.sock:/var/run/docker.sock
 
   backend:
-    image: 'clowder/clowder2-backend:2.0.0-beta.2'
+    image: 'clowder/clowder2-backend:release-v2.0-beta-3'
```
longshuicy, line 40: We haven't officially released yet hence if you wish to use beta3 you have to use this image.
But once released, the tag change will be a part of the release PR.

PR #1167, opened 2024-08-13
fix tests that are failing

tcnichol:
Based on what Sandeep posted on slack, all that we might need to fix the tests are changing docker-compose to docker compose.


longshuicy (approved):
Tests passes now. Thanks for the quick turnaround.

PR #1170, opened 2024-08-20
set workers based on recommended number of cores

longshuicy:
Update:
To avoid the `Namespace already exists`, I turned off the `recreate_views` flag which prevents Beanie from attempting to recreate views if they already exist. This seems to work. 

---

1. I moved all the synchronous operations into a thread pool using the `run_in_threadpool` method.
   - This approach allows the synchronous tasks to be executed in a separate thread, preventing them from blocking the main asynchronous event loop.
   - You can read more about this approach [here](https://medium.com/@saverio3107/understanding-concurrency-with-fastapi-and-sync-sdks-4b5cb956e8e0).

2. More importantly, I realized we were not utilizing the **maximum number of workers**.
   - Uvicorn has built-in process management capabilities that can leverage multiple processes for better concurrency. You can read more about using workers [here](https://fastapi.tiangolo.com/deployment/server-workers/).
   - To set the number of workers, you can use the `--workers {num_of_workers}` option in Uvicorn.
   - Based on this [recommendation](https://stackoverflow.com/a/72605405), I configured 17 workers.

**Expected behavior:**
- When downloading a large dataset, you can immediately navigate away from the page and still request other endpoints.
- The backend will take some time to prepare the ZIP file.
- Once the ZIP is ready and the streaming begins, the download will start automatically, regardless of which page you are on.
- The browser will handle the download process

### DIFF
--- a/.run/uvicorn.run.xml
+++ b/.run/uvicorn.run.xml
@@ -1,7 +1,6 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="uvicorn" type="PythonConfigurationType" factoryName="Python" nameIsGenerated="true">
     <module name="clowder2" />
-    <option name="ENV_FILES" value="" />
     <option name="INTERPRETER_OPTIONS" value="" />
     <option name="PARENT_ENVS" value="true" />
     <envs>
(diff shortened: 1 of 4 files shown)


tcnichol (approved):
I tested this and it worked. I was able to navigate away after clicking 'download' on a large dataset, and the dataset did download once the zip file was created.

lmarini:
Created #1189 to fix the work around.

PR #1190, opened 2024-08-27
1187 migrate collections and spaces as v2 metadata.

tcnichol:
This is a work in progress for now.

To test this, you will first need to have an admin user in v2, and you will have to add the metadata definitions. I have added a new 'Collection' metadata definition. Right now this just has a field for the collection_name and collection_id, along with fields for parents. To handle nested collections beyond 1 level, or to handle datasets that are in more than one collection, the definition will need to change.

### DIFF
--- a/scripts/migration/migrate.py
+++ b/scripts/migration/migrate.py
@@ -592,7 +592,7 @@ def build_collection_hierarchy(collection_id, headers):
     return children
 
 
-def build_collection_metadata_for_v1_dataset(dataset, user_v1, headers):
+def build_collection_space_metadata_for_v1_dataset(dataset, user_v1, headers):
     dataset_id = dataset["id"]
     dataset_collections = get_clowder_v1_dataset_collections(
         headers=headers, user_v1=user_v1, dataset_id=dataset_id
@@ -667,7 +667,7 @@ def process_user_and_resources(user_v1, USER_MAP, DATASET_MAP):
                 add_file_metadata(file, file_v2_id, clowder_headers_v1, user_headers_v2)
         # posting the collection hierarchy as metadata
         # TODO need to actually post this
-        collection_metadata_dict = build_collection_metadata_for_v1_dataset(
+        collection_space_metadata_dict = build_collection_space_metadata_for_v1_dataset(
             dataset=dataset, user_v1=user_v1, headers=clowder_headers_v1
         )
         migration_extractor_collection_metadata = {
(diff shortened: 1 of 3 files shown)


scripts/migration/migrate.py
```diff
@@ -117,6 +117,110 @@ def add_v1_space_members_to_v2_group(space, group_id, headers):
...
+
+def get_clowder_v1_dataset_collections(headers, user_v1, dataset_id):
+    matching_collections = []
+    endpoint = f"{CLOWDER_V1}/api/collections/allCollections"
+    response = requests.get(endpoint, headers=headers)
+    user_collections = response.json()
+    for collection in user_collections:
+        collection_id = collection["id"]
+        if collection["name"] == "subchild":
+            print("HERE")
```
longshuicy, line 134: Is this print needed?

scripts/migration/migrate.py
```diff
@@ -117,6 +117,110 @@ def add_v1_space_members_to_v2_group(space, group_id, headers):
...
+
+
+def get_clowder_v1_collection_self_and_ancestors(
+    collection_id, self_and_ancestors, headers
+):
+    endpoint = f"{CLOWDER_V1}/api/collections/{collection_id}"
+    response = requests.get(endpoint, headers=headers)
+    self = response.json()
+    if "id" not in self:
+        print("HERE")
```
longshuicy, line 173: same Is this print needed?

scripts/migration/migrate.py
```diff
@@ -117,6 +117,110 @@ def add_v1_space_members_to_v2_group(space, group_id, headers):
...
+    parents = []
+    all_collections_v1_endpoint = (
+        f"{CLOWDER_V1}/api/collections/allCollections?limit=0&showAll=true"
+    )
+    response = requests.get(all_collections_v1_endpoint, headers=headers)
+    all_collections = response.json()
+    for collection in all_collections:
+        collection_name = collection["name"]
+        if collection_name == "ROOT C" or collection_name == "ROOT D":
+            print("C OR D")
```
longshuicy, line 213: What does Root C or Root D mean?

scripts/migration/migrate.py
tcnichol: will remove, i was using this print in debugging

scripts/migration/migrate.py
tcnichol: oops. i was checking what happened some specific collections from v1, removed.

tcnichol:
I think I've addressed all comments and everything is working. Got rid of all the print statements I had accidentally checked in for debugging and did some better logging for space and collection.

For posting the space metadata for a dataset, I added a try/except so if we're using v1 without my fix to return the space author, we will post space metadata based on the fields that we have.

PR #1199, opened 2024-10-11
Fix/beta 3 extractions list

lmarini:
Several fixes required for beta-3 release, mostly in the listing of extractor events. This also cleans up a lot of code and combines useEffects calls in one call instead of many. Now the list of extraction events should be scoped to the dataset or file id currently being looked at, instead of showing all events. Also worth noticing, extraction messages were not working due to a change in imports. Please see comments on commits for a full list.


tcnichol:
@lmarini I am taking a look. Looks like the extraction history is fixed, checking over the code now.

PR #1205, opened 2024-10-24
Fix refresh token issue when using password flow

GalMunGral:
- Related:  ncsa/idot-pma#173
- `POST /login` and `POST /auth/login` have the exact same logic. @longshuicy suggested deprecating the `POST /login` endpoint.
- Applied the same logic for refresh tokens as used in the authorization code flow (`GET /auth`) so that access tokens obtained through the password flow (`POST /auth/login`) can also be refreshed by `GET /auth/refresh_token`.

### DIFF
--- a/backend/app/routers/authentication.py
+++ b/backend/app/routers/authentication.py
@@ -8,6 +8,7 @@ from app.keycloak_auth import (
 )
 from app.models.datasets import DatasetDBViewList
 from app.models.users import UserDB, UserIn, UserLogin, UserOut
+from app.routers.utils import save_refresh_token
 from beanie import PydanticObjectId
 from fastapi import APIRouter, Depends, HTTPException
 from keycloak.exceptions import (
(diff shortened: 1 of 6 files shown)


longshuicy (changes requested):
Code looks good and tested. 
One thing I noticed is the "deprecated" not seems to work? I can still POST to the /api/v2/login endpoint with no issue. And i didn't see any notification either.
Could you look into that? if it's not straightforward, maybe we can just delete this endpoint. But we might be careful, a quick search I found there are some scripts (in the scripts folder) still uses this endpoint.

longshuicy:
After we discussed, could you try below:

1. Make a utility function of that tokenDB logic. You can put it in `backend/app/routers/utils.py`.
2. Then revert the deprecate. We might want to keep both endpoints. (so you don't need to worry about making deprecating work)
3. Put the logic in /auth/login and /login

max-zilla:
@longshuicy i will take a look at this also but it's ready for another look, thanks.

longshuicy (approved):
Code looks good and functionality works. Approve.

PR #1206, opened 2024-11-04
Add endpoint for updating user info

GalMunGral:


lmarini (changes requested):
@GalMunGral @max-zilla this works great. I could even change the email and log in with the new email address. Unfortuantely, most of our queries that filter on current user use email and the old email is on the existing resources, resulting in things like listing my datasets not working anymore. For now can we disable changing the email address? We can address the queries later and reanable it then. How does that sound? Thanks!

lmarini (approved):

PR #1232, opened 2025-03-20
1225 integrate mirador js library

sandeep-ps:


frontend/src/components/visualizations/WSI/WSI.tsx
```diff
@@ -0,0 +1,49 @@
...
+			type: "mosaic",
+			height: 1000,
+		},
+		workspaceControlPanel: {
+			enabled: false,
+		},
+		windows: [
+			{
+				manifestId:
+					"http://localhost:8002/iiif-presentation-server/manifest.json",
```
sandeep-ps, line 37: @lmarini, Please change the manifest URL here to the one you need.

frontend/src/app.config.ts
```diff
@@ -88,7 +88,7 @@ config["eventListenerJobStatus"]["skipped"] = "SKIPPED";
 config["eventListenerJobStatus"]["resubmitted"] = "RESUBMITTED";
 
 config["streamingBytes"] = 1024 * 10; // 10 MB?
-config["rawDataVisualizationThreshold"] = 1024 * 1024 * 10; // 10 MB
+config["rawDataVisualizationThreshold"] = 1024 * 1024 * 1024; // 1 GB
```
lmarini, line 91: Can we make the default smaller?

frontend/src/components/visualizations/WSI/manifest.json
```diff
@@ -0,0 +1,18 @@
+{
+	"name": "wsi-visualization",
```
lmarini, line 2: Can we rename this to something more iiif generic? Since any image can be handled this way?

lmarini (changes requested):

frontend/src/app.config.ts
sandeep-ps: Yes, I had planned to keep the previous default as is. This was accidentally committed. I will revert this change.

frontend/src/components/visualizations/WSI/manifest.json
sandeep-ps: Yes, makes sense. I will change it to use IIIF instead.

sandeep-ps: I'm also updating the visualization config MIME type to a custom one image/x-iiif. Please let me know if it needs to be kept as application/octet-stream.

PR #1252, opened 2025-11-18
enable integration with external keycloak clients

hwenchi:
## Summary

- Adds optional `redirect_uri` query param to the existing `GET /auth/token` endpoint. If omitted, defaults to Clowder's own `settings.auth_redirect_uri`.
- Refactors `get_token` in `utils.py` to accept `redirect_uri` as a keyword argument, using the module-level `keycloak_openid` instance directly.

## Background

IDOT TREC shares the same Keycloak realm and client (`clowder2-backend`) as Clowder. When a user logs in via IDOT TREC's login button, Keycloak redirects back to the **IDOT TREC frontend** with an authorization code. IDOT TREC's backend must then call Clowder's `/auth/token` to exchange that code — not just hit Keycloak directly — because `/auth/token` upserts the user record into Clowder's MongoDB, which IDOT TREC relies on for group membership and permissions.

**Why `redirect_uri` must be configurable:**

Keycloak requires that the `redirect_uri` passed during the token exchange exactly matches the one used in the original authorization request. Since IDOT TREC's login flow redirects to the IDOT TREC frontend (not Clowder's backend), IDOT TREC must pass its own frontend URL when calling `/auth/token`. See [ncsa/idot-pma@login-oidc: clowder/clowder.py#L305-310](https://github.com/ncsa/idot-pma/blob/login-oidc/backend/clowder/clowder.py#L305-L310).

## Used by
**ncsa/idot-pma#213** — calls `/auth/token` passing its own `redirect_uri` to support the OIDC authorization code flow from an external frontend.


longshuicy (approved):
The /token endpoint is currently only used by IDOT TREC, but it's designed to support similar auth flows in the future; so if other integrations need the same pattern, we can point them there as well.

